PR #5, opened 2020-02-28
modified file paths and variable declarations

bdolor:
- `TEMPLATES_FILE` was pointing to a non-existent file.
- `cur_time` was being referenced before it was declared. Underscores added for better file naming. 
- `OUTPUT` was being saved as `.txt` when output is `.json`. Location of saved data file moved to `/data` directory.
- `fake_pii_csv` was pointing to a non-existent file.


omri374:
Thanks for the PR. I'll review the rest of them in the next few days.

PR #11, opened 2020-03-28
Update replacement for Flair evaluation

kirkins:
I noticed during the notebook `Evaluate flair models` the first 2 models `ner` and `fast-ner` don't produce results for `PERSON` or `GPE`.

When looking at the evaluation code I noticed the line:

    new_tags.append("PERSON" if tag == "PER" else tag)

This made me realize the problem was due to the model using different labels. I believe the intention of the line above was to replace the `PER` style tags with `PERSON`. However it never matches any tags because they are always start with a letter and dash for example `E-PERSON` or `S-PERSON`.

To fix this I replaced the `==` comparison with a regex based comparison and did the same thing with `LOC` which is expected to be `GPE` by the evaluator. Tested with both the 4 class NER and 18 class NER to make sure the regex only matches `PER` and not `PERSON`.

### DIFF
--- a/presidio_evaluator/flair_evaluator.py
+++ b/presidio_evaluator/flair_evaluator.py
@@ -1,4 +1,5 @@
 from typing import List
+import regex
 
 try:
     from flair.data import Sentence, build_spacy_tokenizer
@@ -69,6 +70,13 @@ class FlairEvaluator(ModelEvaluator):
 
         new_tags = []
         for tag in tags:
-            new_tags.append("PERSON" if tag == "PER" else tag)
+            #new_tags.append("PERSON" if tag == "PER" else tag)
+            is_person = regex.compile('[A-Z]-PER\\b')
+            if is_person.match(tag):
+                tag = tag.replace("PER", "PERSON")
+            is_gpe = regex.compile('[A-Z]-LOC\\b')
+            if is_gpe.match(tag):
+                tag = tag.replace("LOC", "GPE")
+            new_tags.append(tag)
 
         return new_tags
--- a/presidio_evaluator/flair_evaluator.py
+++ b/presidio_evaluator/flair_evaluator.py
@@ -70,7 +70,6 @@ class FlairEvaluator(ModelEvaluator):
 
         new_tags = []
         for tag in tags:
-            #new_tags.append("PERSON" if tag == "PER" else tag)
             is_person = regex.compile('[A-Z]-PER\\b')
             if is_person.match(tag):
                 tag = tag.replace("PER", "PERSON")


omri374:
Hi @kirkins 
Looks like there was an update on Flair. In any case, does the `_to_io` method on `ModelEvaluator` help?
https://github.com/microsoft/presidio-research/blob/0c043922a85dcf4ea6c30923f445204cc442269b/presidio_evaluator/model_evaluator.py#L116

If yes, it's on the super class of FlairEvaluator, so you can just call it.

kirkins:
I don't think it fixes the issue I was looking at. In the notebook `Evaluate flair models` the `ner` and `ner-fast` models don't return results for `GPE` and `PERSON` as shown below:

Screenshot from 2020-04-04 14-00-05

After the change those same models show results below:

Screenshot from 2020-04-04 13-55-03

_While testing I had to re-run `python3 setup.py install` for the change to take effect._

omri374:
I see. There some helper functions for translating entity names from one dataset to the other. 
On the `FlairEvaluator` there's a call to `sample.translate_input_sample_tags` which can be turned on and off. In addition, you can call `sample.translate_tags` with your own mapping dictionary to replace entity names on the original sample with the ones returned by the model in hand.

https://github.com/microsoft/presidio-research/blob/0c043922a85dcf4ea6c30923f445204cc442269b/presidio_evaluator/data_objects.py#L389

The problem with your proposed solution, is that it would work for the pretrained Flair models, but wouldn't for a Flair model trained from scratch, potentially with other entity names.

omri374:
Regarding the `python3 setup.py install` note, you can try `pip install -e .` from the root of the project. This installs the package in editable model, and then you can change code while running it on Jupyter notebooks (if that's the setting you have)

kirkins:
Thanks, I think this `translate_to_spacy_entities` function runs by default. So I try adding this to the default dictionary, `PRESIDIO_SPACY_ENTITIES`:

    "PER": "PERSON",

But it doesn't seem to fix the results for `PERSON`. Would you expect that to work?

omri374:
The `sample.translate_input_sample_tags()` call can replace any entity names in the input sample with anything else, given a dictionary. Assuming it's working, I believe it should answer your need. If there's a bug somewhere, let's fix it. Can you perhaps write a pytest test to verify this? I'd prefer a generic solution (assuming there's a bug) over a hotfix for the FlairEvaluator only.

kirkins:
That sounds like a good plan. If you don't hear from me for some time it's because I've been quarantined without my laptop. I'd like to take a crack at writing the pytest once I have computer access. All the best, Philip.

omri374:
@kirkins closing for now, please feel free to reopen if you have additional changes you wish to introduce.

PR #27, opened 2021-05-18
I2B2 data formatter

navalev:
convert I2B2 data (2014 version) to InputSample structure

### DIFF
--- a/presidio_evaluator/dataset_formatters/i2b2_formatter.py
+++ b/presidio_evaluator/dataset_formatters/i2b2_formatter.py
@@ -11,10 +11,10 @@ import xmltodict
 from presidio_evaluator import InputSample
 from presidio_evaluator.dataset_formatters import DatasetFormatter
 
-class I2C2Formatter(DatasetFormatter):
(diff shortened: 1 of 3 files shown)


presidio_evaluator/dataset_formatters/i2b2_formatter.py
```diff
@@ -0,0 +1,55 @@
...
+import xmltodict
+
+from presidio_evaluator import InputSample
+from presidio_evaluator.dataset_formatters import DatasetFormatter
+
+class I2B22014Formatter(DatasetFormatter):
+  def __init__(
+        self,
+        files_path=Path("../../data/i2b2").resolve(),
+        glob_pattern: str = "*.*",
```
omri374, line 18: If you're not using glob then this is probably not needed.

omri374: We'll fix it later on

navalev: forgot to remove it

PR #36, opened 2021-10-11
Adding a generator class that leverages faker parse

prvenk:
Focus is essentially to parse a template file and return a list of fake PII sentences with tokens replaced by faker examples. 

Doesn't work on most token categories, since token names are different. Works on a few tokens like address if I do a regex and change token from uppercase to lowercase. Added a dummy print whenever there is an exception.

If the `parse` function is extended and original template files changed to be in line with faker token names, then this approach may look leaner.

Also, the constructor takes in a `locale`. Depending on context of the template, that may be useful. Default locale is `en_us`.

@omri374 Thoughts?


prvenk:
@omri374 made all changes and added new faker based templates....review once more?

omri374:
Looks awesome! Thanks

PR #47, opened 2022-07-20
Add Presidio packages required dependencies

Robbie-Palmer:
- The `presidio_anonymizer` package is used by `presidio_pseudonymize`
- The `presidio_analyzer` package is used by `presidio_pseudonymize`, `scorers` and `models`


omri374:
Thanks! good catch

omri374 (approved):

PR #50, opened 2022-08-05
PresidioSentenceFaker

Robbie-Palmer:
`1_Generate_data.ipynb` and `presidio_data_generator.py`'s `main` function contain code for generating fake records utilising the new providers and the mappings defined in this package.
This is a useful higher-level API for consuming the functionality of this package in other projects e.g.

```python
from presidio_evaluator.data_generator import PresidioFakeRecordGenerator

entity_generator = PresidioFakeRecordGenerator(locale="en_US", lower_case_ratio=0.05,
                                               random_seed=42)
fake_patterns = entity_generator.generate_new_fake_records(num_samples=10000)
```

The version of the code in this PR takes from the notebook, which includes a mapping from Faker entity types to Presidio entity types, which isn't done in `presidio_data_generator.py`'s `main` function.
Otherwise, it is the same code, just wrapped up into a class.

In this PR, I've also changed what I think are bugs.
- The `presidio_entities_map` which maps Faker entity types to Presidio  entity types
    - Included `"URL": "DOMAIN_NAME"` which I believe should be the other way round
    - Included `"DOMAIN_NAME": "DOMAIN_NAME"` but `DOMAIN_NAME` isn't listed in https://microsoft.github.io/presidio/supported_entities/
- `PresidioAnalyzerWrapper`'s `predict` method is hard coded to always use `"en"` instead of the provided language
- `PresidioAnalyzerWrapper`'s construction logic around language is confusing, especially between providing an analyzer or not
    - Validated that if an analyzer is provided, the chosen language must be supported by the analyzer
    - If an analyzer is not provided, construct a new one, passing in the language as the only supported language

And made some stylistic changes which I think makes the code more readable, though obviously, this is subjective
- Dict literals to dict constructors
- Usage of `/` for creating pathlib Paths
- Usage of f strings
- Formatting from PyCharm, just because this is the IDE, I'm using, and it makes it easy to keep things tidy
    - If different formatting is preferred, it may be helpful to configure it using git pre-commit hooks, black and flake8, as outlined here: https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/
        - I'm happy to do this if you think it is useful


omri374:
Thanks @Robbie-Palmer. We'll review this soon.

omri374:
Thanks @Robbie-Palmer, this is a great addition and greatly improves the codebase! 
A few mimor comments:
1. Could you add a few unit-tests for the new class?
2. The distinction between the previous and suggest class mighit not be 100% clear to the user. Should we update the tutorials or add more clarification? Perhaps rename it to something else? (I don't have a suggestion, just a thought).
3.

presidio_evaluator/data_generator/faker_extensions/providers.py
```diff
@@ -7,15 +7,16 @@
 from faker.providers import BaseProvider
 from faker.providers.address.en_US import Provider as AddressProvider
 from faker.providers.phone_number.en_US import Provider as PhoneNumberProvider
+from presidio_evaluator.data_generator import raw_data
+
+_raw_data_dir = Path(raw_data.__path__[0])
```
omri374, line 12: This is causing an issue with python 3.7: https://dev.azure.com/csedevil/Presidio/_build/results?buildId=14436&view=logs&j=00b6a206-eb91-585b-3d14-cf1a4d7b1970&t=9a9400f1-b707-5d54-e2fd-4fe1d8bc6dda

similar case: https://code.djangoproject.com/ticket/21862

Robbie-Palmer:
@omri374 I'm sorry for the really long delay! I've been pulled into a number of unrelated projects these past few months
No availability this coming week, but I'll prioritise addressing your comments asap afterwards

Robbie-Palmer:
Hi @omri374, I've fixed the issue with Python 3.7, added a tes, added to the docs and updated the Generate data notebook

Wrt the naming, my instinct would be to rename `PresidioDataGenerator` to `FakeSpanGenerator`.
- Dropping `Presidio from the front would make it clearer that it doesn't return `Presidio` entity-types but rather `Faker` entity-types
- Changing `Data` to `FakeSpan` makes it less generic and highlights the difference between this and calling `Faker` directly (the spans)

Or would be great if the name could signal the difference between it and `Faker` being the semantic dependency maintenance across sentences, but I don't have any ideas around that.
Though I didn't want to rename it and break the current API
Do you have any ideas/preferences?

On updating this, I found and fixed some other bugs that were introduced since I opened this PR
- Added missing `religions.csv` and `us_driver_license_format.yaml` to package
- Added `ReligionProvider` to `faker_extensions/__init__.py`
- `UsDriverLicenseProvider` had changed its method name from `us_driver_license` to `driver_license` causing the templates to break. I changed it back

I also merged `master` into this branch to resolve conflicts
I don't know why GitHub is now showing the commits I pulled in as if they're unique to my branch in the diff, since I'm pointing against `master` which has these included
E.g. It's displaying as though I'm adding `.pipelines/publish-to-pypi.yml` when this [exists in the master branch](https://github.com/microsoft/presidio-research/blob/master/.pipelines/publish-to-pypi.yml) :man_shrugging:

.pipelines/publish-to-pypi.yml
```diff
@@ -0,0 +1,33 @@
...
+    inputs:
+
+      pythonUploadServiceConnection: Pypi-Presidio
+  - task: Bash@3
+    displayName: 'Upload Wheel to PyPi'
+    inputs:
+      targetType: 'inline'
+      script: |
+        set -eux  # fail on error
+        twine upload -r presidio --verbose --config-file $(PYPIRC_PATH) --skip-existing dist/*
```
Robbie-Palmer, line 33: The pipeline is failing because `$(PYPIRC_PATH)` is empty
How can this be set? And why is it set for other branches but not this one?

omri374: This is the release pipeline, it is not meant to be ran locally. The PYPIRC_PATH is the path to the configuration for releasing to PyPI, which is only exposed to administrator.

Robbie-Palmer: It's being triggered on Azure Pipelines and failing
When I assume it shouldn't be running that step

https://dev.azure.com/csedevil/Presidio/_build/results?buildId=14665&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=dee4e6ec-be1b-5671-4678-ab293b4004c5&l=23

omri374: You're right. We've disabled it.

omri374:
@Robbie-Palmer Will first rebase your fork's master to this repo's master, and then merge the fork master into this feature branch, help?

Robbie-Palmer:
@omri374 the PR diff is now showing only the changes from this branch
I edited the PR to point to another branch then edited it to point back to master which forced GitHub to reload the diff
https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -295,48 +292,119 @@ def name_gendered(row):
...
+
+        self.fake_records = None
+
+    def generate_new_fake_records(self, num_samples: int) -> List[FakerSpansResult]:
+        self.fake_records = list(self._data_generator.generate_fake_data(templates=self._sentence_templates,
+                                                                         n_samples=num_samples))
+        # Map faker generated entity types to Presidio entity types
+        for sample in self.fake_records:
+            for span in sample.spans:
+                span.type = self.faker_to_presidio_entity_type[span.type]
```
melmatlis, line 396: Perhaps we can add to the faker_to_presidio mapping an additional default value 'other' for cases when a new unmapped key appears, then the mapping will become. What do you think?
```suggestion
                span.type = self.faker_to_presidio_entity_type.get(span.type,'other')
```

melmatlis (commented):
Thank you @Robbie-Palmer for the PR.
You've introduced some clear coding and pretty neat refactoring to the code.
One small general comment, I think using the names `PresidioFakeRecordGenerator` and `PresidioDataGenerator` are a bit confusing, perhaps we can speak with @omri374 and try to think of more distinguished naming

omri374 (commented):
First of all, thank you @Robbie-Palmer! this PR introduces really great code improvements and bug fixes.

On the new class topic, I agree we should first align on the right structure. Backward compatibility is not a big issue IMHO as this package was just recently released to PyPI and we can create a deprecation notice for the old class name.

Here are the options as I see it:
1. Continue having one class, and using defaults to simplify the usage
2. Continue having one class with two methods for simple and customized use
3. Have two classes, which is proposed in this PR
4. Have a base class, which is inherited by another class, similar to this PR but maintaining some relation between the two.

Another discussion was on the name(s) of classes. 

My preference is to have one class, with a refined name, which would use defaults whenever a user is interested in simple generation.

Happy to hear your thoughts @Robbie-Palmer and @melmatlis. 
We could also set up a short call if that would be the easiest.

presidio_evaluator/data_generator/faker_extensions/providers.py
```diff
@@ -69,13 +67,11 @@ def company(self):
...
         super().__init__(generator=generator)
-        us_driver_license_file = Path(
-            Path(__file__).parent.parent, "raw_data", "us_driver_license_format.yaml"
-        ).resolve()
+        us_driver_license_file = (raw_data_dir / "us_driver_license_format.yaml").resolve()
         formats = yaml.safe_load(open(us_driver_license_file))
         self.formats = formats['en']['faker']['driving_license']['usa']
 
-    def driver_license(self) -> str:
+    def us_driver_license(self) -> str:
```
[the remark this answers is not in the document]
melmatlis: As long as compatibility is not an issue, I'm all good :)

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -295,48 +292,119 @@ def name_gendered(row):
...
+
+        presidio_additional_entity_providers = [IpAddressProvider,
+                                                NationalityProvider,
+                                                OrganizationProvider,
+                                                UsDriverLicenseProvider,
+                                                AgeProvider,
+                                                AddressProviderNew,
+                                                PhoneNumberProviderNew,
+                                                ReligionProvider]
+        if additional_entity_providers is None:
```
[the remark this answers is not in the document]
Robbie-Palmer: Yeh this caught me off, I originally set the default argument to an empty list
But the tests caught that if you have this module imported, and create multiple instances of this class, altering `additional_entity_providers` for any, alters the list for all instances

Turns out this is a very [unintuitive aspect of Python's design](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument) where the empty list is attached to the function definition, not to the class instance

So the recommended approach for mutable default arguments is to set them to `None` and explicitly check `is None` https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

presidio_evaluator/data_generator/faker_extensions/providers.py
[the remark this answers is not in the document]
Robbie-Palmer: Yeh this is rolling back the change in this commit https://github.com/microsoft/presidio-research/commit/e0dfd585d8bdafa8a472c8500e328e35ad325b29
Which changed the name of the property from `us_driver_license` to `driver_license` which broke running with the existing template

The new tests that use the template would now catch this, so can make changing this easier

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -122,7 +125,7 @@ def parse(
     def read_template_file(templates_file):
         with open(templates_file) as f:
             lines = f.readlines()
-            lines = [line.replace("\\n", "\n") for line in lines]
+            lines = [line.strip() for line in lines]
```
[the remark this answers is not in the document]
Robbie-Palmer: No, The templates currently in there don't have `\\n` in them
I'm not sure what scenario this is accounting for? Is there a known dataset that this needs done for?
`\\n` would imply the writer is intentionally escaping the escape character to avoid the new line from existing

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -295,48 +292,119 @@ def name_gendered(row):
...
+                                         state_abbr="GPE",
+                                         age="AGE")
+
+    def __init__(self,
+                 locale: str,
+                 lower_case_ratio: float,
+                 additional_entity_providers: List[BaseProvider] = None,
+                 additional_sentence_templates: List[str] = None,
+                 random_seed: SeedType = None):
+        presidio_templates_file_path = raw_data_dir / "templates.txt"
```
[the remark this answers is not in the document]
Robbie-Palmer: With the current API, a user can load the sentences in before constructing this class, and pass them in as a list of strings to the `additional_sentence_templates` parameter e.g.
```python
my_custom_templates = PresidioDataGenerator.read_template_file('~/my_templates.txt')
my_custom_templates.extend(PresidioDataGenerator.read_template_file('~/other_dir/more_templates.txt'))

record_generator = PresidioFakeRecordGenerator(locale='en', lower_case_ratio=0, additional_sentence_templates=my_custom_templates)
```

notebooks/1_Generate_data.ipynb
```diff
@@ -201,13 +162,16 @@
...
    "outputs": [],
    "source": [
-    "fake.add_provider(IpAddressProvider)  # Both Ipv4 and IPv6 IP addresses\n",
-    "fake.add_provider(NationalityProvider)  # Read countries + nationalities from file\n",
-    "fake.add_provider(OrganizationProvider)  # Read organization names from file\n",
-    "fake.add_provider(UsDriverLicenseProvider)  # Read US driver license numbers from file\n",
-    "fake.add_provider(AgeProvider)  # Age values (unavailable on Faker)\n",
-    "fake.add_provider(AddressProviderNew)  # Extend the default address formats\n",
-    "fake.add_provider(PhoneNumberProviderNew)  # Extend the default phone number formats"
+    "from presidio_evaluator.data_generator.faker_extensions.providers import *\n",
```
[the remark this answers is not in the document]
Robbie-Palmer: It runs ok, the classes are imported, and these lines are just statements of the class names
The output of the cell is the final class turned into a string for the notebook output, so `presidio_evaluator.data_generator.faker_extensions.providers.ReligionProvider`

presidio_evaluator/data_generator/presidio_data_generator.py
Robbie-Palmer: It makes sense to me! Though would also need to change a bit more than this one line, as we need to track these unknown types and replace them in the template too
I would probably set this as another feature introduced in a new PR with additional tests, even though it's a relatively small change

presidio_evaluator/models/presidio_analyzer_wrapper.py
```diff
@@ -66,65 +68,53 @@ def predict(self, sample: InputSample) -> List[str]:
...
 
         if "ORGANIZATION" in self.entities and "ORGANIZATION" not in supported_entities:
             recognizers = analyzer_engine.get_recognizers()
-            spacy_recognizer = [
-                rec
-                for rec in recognizers
-                if rec.name == "SpacyRecognizer" or rec.name == "StanzaRecognizer"
-            ]
+            spacy_recognizer = [rec for rec in recognizers
+                                if rec.name == "SpacyRecognizer" or rec.name == "StanzaRecognizer"]
```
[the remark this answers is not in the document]
Robbie-Palmer: Support for this isn't currently in `master`, should this be fixed as a separate PR for supporting `presidio` v2.2.30 when this was introduced? Along with any other API changes that may have happened?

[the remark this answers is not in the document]
Robbie-Palmer: My thoughts being that the other bug fixes I included in this PR were directly triggered by this new class and the tests
Updating `PresidioAnalyzerWrapper` to support `TransformersRecognizer` would ideally come with additional tests targeted at this other class

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -295,48 +292,119 @@ def name_gendered(row):
...
-    sentence_templates = PresidioDataGenerator.read_template_file(template_file_path)
-    fake_patterns = data_generator.generate_fake_data(
-        templates=sentence_templates, n_samples=10000
-    )
-
-    # save to json
-    output_file = Path(
-        Path(__file__).parent.parent.parent, "data", "presidio_data_generator_data.json"
-    )
+class PresidioFakeRecordGenerator:
```
Robbie-Palmer, line 295: @omri374 @melmatlis
Easier to have a thread here to discuss the naming of this class and `PresidioDataGenerator` (and can have a follow-up call as @omri374 suggested)

My instinct is to rename both classes.

Wrt `PresidioDataGenerator` I find it surprising that it returns `Faker` entity types; as its name suggests, it'll return data in the "Presidio" form.
I also think the name could be made more concrete, as the term "Data" is very general, not giving a sense of what will be returned
I think the key features of this class, that could be captured in its name are:
- It is a wrapper around Faker, extending its functionality
- It returns `FakeSpan`s
- It works with `templates` - Critically so that it can maintain semantic dependency across sentences

I'd maybe even go with something as simple as `SentenceFaker`
Which captures the difference between this and `Faker` and when you would want to use one versus the other

Wrt `PresidioFakeRecordGenerator`, the key features of this class vs the other are:
- Uses `presidio` not `faker` entity-types
- Defaults to using all additional `presidio` entity providers and templates

So could be `PresidioSentenceFaker`
Distinguishing how `SentenceFaker` is in the `faker` domain and this class is the same idea but in the `presidio` domain

What are your thoughts?

presidio_evaluator/data_generator/presidio_data_generator.py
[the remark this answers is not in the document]
melmatlis: Interesting, thank you for sharing that info with me.

notebooks/1_Generate_data.ipynb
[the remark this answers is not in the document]
melmatlis: For me, using VScode, it failed.

presidio_evaluator/data_generator/presidio_data_generator.py
melmatlis: makes sense

presidio_evaluator/data_generator/presidio_data_generator.py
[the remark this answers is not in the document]
melmatlis: Not sure, but I would keep the line just for the continuity.

presidio_evaluator/models/presidio_analyzer_wrapper.py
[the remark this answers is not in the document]
melmatlis: @omri374 , the `TransformersRecognizer` is just in the samples, not part of the repo itself

presidio_evaluator/data_generator/presidio_data_generator.py
melmatlis: I suggest we have a call on this topic. 
Ideally I would like to design a single class which enables users to:
1. Use Faker as well as custom providers
2. Use default as well as user provided templates
3. Enable mappings between template <-> faker <-> presidio entity names

presidio_evaluator/models/presidio_analyzer_wrapper.py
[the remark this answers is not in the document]
omri374: I meant this one, which is used to capture the output of the `TransformersNlpEngine`
https://github.com/microsoft/presidio/blob/main/presidio-analyzer/presidio_analyzer/predefined_recognizers/transformers_recognizer.py

presidio_evaluator/data_generator/presidio_data_generator.py
omri374: Let's have a call. @Robbie-Palmer could you please email presidio@microsoft.com and we'll set up a call?

presidio_evaluator/data_generator/presidio_data_generator.py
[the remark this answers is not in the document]
Robbie-Palmer: I've been viewing this class as a high-level API which, by default, will use all the power/functionality provided by this library. Just as the code that already existed in the main function and the notebook was there as what a layman would run
Primarily to target the ML Engineer persona who wishes to evaluate the configurable models for generic PII detection, rather than the Data Scientist persona who wants to build their model / drop lower level into configuring templates/providers
If they wish to strip back by using fewer templates or fewer providers, that's when they drop to the lower-level API

Though it would be nice to be able to easily choose to extend or override, at the moment, you can only override
One option would be a flag parameter for extending or overriding, but I think that makes a clunky API
Or another option is defaulting to the existing templates and providers, and override these with those provided instead of extending. I think it makes sense to provide this default for ease of use, and it doesn't make sense to default to nothing as then the constructed object is unusable. If overriding instead of extending, then just need to make it easy to load all existing templates/providers prior to constructing the class to accommodate that use case

notebooks/1_Generate_data.ipynb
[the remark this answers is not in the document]
Robbie-Palmer: What error are you getting?
It runs for me in both Jupyter Lab and VS Code

presidio_evaluator/data_generator/presidio_data_generator.py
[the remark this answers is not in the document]
Robbie-Palmer: I've added `strip()` which removes new-line characters at the start/end of the line since these are artefacts of being a file
If I retain `replace("\\n", "\n")` then if `\\n` appears at the start of end of the line it will now be removed
This feels strange to me, as the extra escape character suggests this is there intentionally as part of the template and not an artefact of the encoding
Happy to re-include it if you're happy that `\\n` strings at the start/end of of lines will be removed

presidio_evaluator/data_generator/presidio_data_generator.py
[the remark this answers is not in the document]
Robbie-Palmer: I updated the API to override instead of extending, to support not wanting to use the inbuilt templates/providers
https://github.com/microsoft/presidio-research/pull/50/commits/407718d1827751a0f2116a633dc2e00b4cf5fbfe

presidio_evaluator/data_generator/presidio_data_generator.py
Robbie-Palmer: Opened a new issue to track this https://github.com/microsoft/presidio-research/issues/68

presidio_evaluator/models/presidio_analyzer_wrapper.py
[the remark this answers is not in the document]
Robbie-Palmer: I've opened a separate issue to track this https://github.com/microsoft/presidio-research/issues/69

presidio_evaluator/data_generator/presidio_data_generator.py
Robbie-Palmer: @omri374 and @melmatlis to summarise what we discussed about this PR in the call
The current `PresidioDataGenerator` API looks like:
- Construct a `Faker` object and pass to constructor
- Choose locale and lower case ratio
- Generate faker spans (with faker entity types) from a list of templates
- Parse a single template string 
- `add_provider_alias`

The API in this PR looks like:
- Choose your
    - Locale
    - Lower case ratio
    - Entity providers
    - Sentence templates
    - Aliases (via a class field rather than via the constructor)
- Default to using all providers and templates within this repo
- Generate faker spans with presidio entity types

The fact that the `PresidioDataGenerator` takes a `Faker` object into its constructor and returns `Faker` entity types is why I think of it as being in the "Faker domain"
The new API in this PR doesn't expose that `Faker` is being used at all, keeping it in the "Presidio domain", which ties in with what @melmatlis said about wanting to expand to using other libraries with similar but extended functionality

So the differences between the two APIs at the moment are:
- The original requires construction of a `Faker` object by the caller
    - I don't know what use cases there may be for this, which is what I was referring to when saying perhaps researchers want this lower level API? But maybe they don't need it, if there are plans to abstract away the word faking interface to enable use of other libraries too
-  Returning faker entity types vs presidio entity types
- Original can parse a single template string, while new API cannot atm (can easily be added) 

So the questions I have are:
- Is there a use case where a user of this library wants to construct their own `Faker` instance and have `Faker` entity types returned?
    - If yes, should this be pulled into an explicitly `Faker` library oriented module, especially if there are plans to support alternative libraries? And maybe even pushed upstream to `Faker` itself in the long term?
    - If no, then are we happy to deprecate the old API and use the new one instead, therefore melding into one class?

omri374: Hi @Robbie-Palmer. Thanks for the great summary. Initially there was an ambition to expose a `Faker` functionality to make it easy to use Faker objects, but I agree that this is not needed.

My suggestion is to proceed with the two classes approach (which is what this PR currently proposes, correct?), where one is more of the API and the other is wrapping Faker. The Faker wrapper isn't something that someone would typically use, and it further could be replaced if someone wants to use a different fake sentence generator.

If we go with two classes then with regards to naming the classes and the abstraction of Faker, we can consider calling the Faker wrapper class something with `Faker` in it. Maybe SpanFaker, TemplateToFakeDataGenerator or FakeSpanGenerator? (Although the span itself isn't fake, just the text... I'm not sure. Naming is always difficult :) )

I've left a few additional comments to still allow further customizability without having to inherit the class or change the code, but open to your thoughts. Thanks!

notebooks/1_Generate_data.ipynb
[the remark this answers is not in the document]
omri374: Perhaps we can have them as part of the import instead of using `*` and then having the class objects explicitly. Would that work?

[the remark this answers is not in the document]
Robbie-Palmer: @omri374 are you getting an error when you try to run it?

presidio_evaluator/data_generator/presidio_data_generator.py
```diff
@@ -273,48 +280,108 @@ def name_gendered(row):
...
+                                         nation_plural="NRP",
+                                         first_name_nonbinary="PERSON",
+                                         postcode="STREET_ADDRESS",
+                                         secondary_address="STREET_ADDRESS",
+                                         job="TITLE",
+                                         zipcode="ZIP_CODE",
+                                         state_abbr="GPE",
+                                         age="AGE")
+
+    def __init__(self,
```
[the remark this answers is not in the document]
Robbie-Palmer: I've made setting the PII records customisable at the instance level 👍 
And I've [added to the README](https://github.com/Robbie-Palmer/presidio-research/tree/fake-record-generator/presidio_evaluator/data_generator#example) instructions for adding additional entity type mappings
> Add a mapping from the output provider entity type to a Presidio recognised entity type where appropriate
e.g. For a TownProvider which outputs entity type of town, execute PresidioSentenceFaker.ENTITY_TYPE_MAPPING['town'] = 'GPE')

I've left the mapping at the class level, I don't think it needs to be available via the constructor for the instance level?

presidio_evaluator/data_generator/presidio_data_generator.py
Robbie-Palmer: I was trying to avoid doing a large refactor in case this became one of those PRs that lives too long and doesn't get merged 😅 
But I have went for it now
I've removed the overlap that was between the classes, renamed them and moved what remains of the old one into the `faker_extensions` package, consolidating it into a module with related classes

omri374: Thank you! We didn't mean for this to be too big. Appreciate all the work!

Robbie-Palmer: No problem! My main concern isn't the amount of work; it's more that the larger the changes, the more subjective it becomes. And the subjectivity of your team is more important than mine since long-term maintenance depends on the maintainers having a shared vocabulary / ubiquitous language
Plus changes I make could clash with internal work/communication I'm not aware of, so is easy to make broad sweeping changes that clash and then slow progress down

Also there's the worry that starting to refactor some things leads to a snowball effect
E.g. now in `PresidioSentenceFaker` I'm importing `FakerSpansResult as FakeSentenceResult` to avoid the "Faker domain" from leaking into this module
But looking at `FakerSpanResult` it could be renamed generally, as nothing about it is `Faker` specific other than it holding `List[FakerSpan]`
But looking at `FakerSpan` nothing about it is `Faker` specific
So both those classes could be renamed to be more generic, and the `data_objects` module that contains them should then be moved out of the `faker_extensions` package

There are also other examples like that, so the other main concern is finding a good cut-off point for this piece of work, and ensuring any further refactorings are done incrementally
Especially so that the bug fixes are delivered sooner rather than later to avoid other users being affected :)

notebooks/1_Generate_data.ipynb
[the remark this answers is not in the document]
melmatlis: I tried to run it again, and this time ran fine. Closing this thread.

presidio_evaluator/data_generator/presidio_data_generator.py
melmatlis: Completely agree Robbie with your last comment.
@omri374, I think you've mentioned you created a feature branch for all the refactoring we're doing on the code.
What if we change the target of the this PR to the feature branch, by this we will achieve 2 things
1. we won't break main 
2. It will enable us to build on top of Robbie's excellent modifications without the "stress" of prolonging the PR
What do you think?

omri374: Yes, pushing everything into a feature branch would be a good idea. It would also allow me to test this locally without having to clone @Robbie-Palmer's fork :)

omri374: Done. Changed the target branch to data-generator-2.1

omri374: I also completely agree. I suggest to merge this into a feature branch, and plan any changes we see that are needed (such as the one you mentioned @Robbie-Palmer in the faker objects) prior to going and implementing them. If you agree, we'll continue the work on the feature branch also as there are other changes happening on the Evaluator side.

Robbie-Palmer: Yes I'm happy with that :)

omri374: Thanks :)

omri374: BTW, this looks great. In my view could be merged to the main branch. We're putting it in a feature branch so that we could keep the consistency with the other parts that are changing (the evaluator part) @Robbie-Palmer @melmatlis @tranguyen221

PR #52, opened 2022-10-26
Update Flair Model Training

benkilimnik:
This PR updates `models/flair_model.py` and `models/flair_train.py` to work with the [newest Flair release](https://github.com/flairNLP/flair/releases/tag/v0.11). 

It also adds the following:
1. `train_with_transformers` to train a flair NER model with transformers
2. `checkpoint_path` parameter to resume training from a saved model 
3. optional `in_memory` parameter for `read_corpus` to stream the flair corpus from disk instead of loading it into memory all at once


benkilimnik:
@microsoft-github-policy-service agree

omri374:
Thanks! great contribution

omri374 (approved):

PR #58, opened 2022-11-22
Scripts for uploading presidio-evaluator to PyPI

omri374:
Adding a pipeline to push presidio-evaluator to PyPI. Reusing the presidio scripts from here: https://github.com/microsoft/presidio/blob/main/.pipelines/templates/publish-to-pypi.yml
Azure Pipeline can be found here: https://csedevil.visualstudio.com/Presidio/_build?definitionId=267&_a=summary


.pipelines/publish-to-pypi.yml
```diff
@@ -0,0 +1,33 @@
+variables:
```
omri374, line 1: taken from https://github.com/microsoft/presidio/blob/main/.pipelines/templates/publish-to-pypi.yml

.pipelines/publish-to-pypi.yml
```diff
@@ -0,0 +1,33 @@
...
+    inputs:
+
+      pythonUploadServiceConnection: Pypi-Presidio
+  - task: Bash@3
+    displayName: 'Upload Wheel to PyPi'
+    inputs:
+      targetType: 'inline'
+      script: |
+        set -eux  # fail on error
+        twine upload -r presidio --verbose --config-file $(PYPIRC_PATH) --skip-existing dist/*
```
melmatlis, line 33: Should this be `twine upload -r presidio-evaluator`?

omri374: That's what I though originally, but -r is the repository, and we have the PyPI setup in AzDO where under the repository "presidio" we the configuration and credentials.

melmatlis (approved):

PR #61, opened 2022-12-06
Create LocalExperimentTracker for storing information in local locations

melmatlis:
The code generates a new directory per experiment and enables storing the following information:
1. Store a confusion matrix as a comma separated file
2. Store a serialized version of model errors for down stream analysis. The file is stored as a Pickle object

### DIFF
--- a/presidio_evaluator/experiment_tracking/local_tracker.py
+++ b/presidio_evaluator/experiment_tracking/local_tracker.py
@@ -95,3 +95,9 @@ if __name__ == "__main__":
     local_tracker.log_confusion_matrix_table(matrix, labels)
 
     local_tracker.end()
+
+    # view stored files
+    print('experiment generated the following files:')
+    experiment_path = Path(directory, experiment_name)
+    for f in experiment_path.iterdir():
+        print(f.as_posix())
--- a/presidio_evaluator/experiment_tracking/local_tracker.py
+++ b/presidio_evaluator/experiment_tracking/local_tracker.py
@@ -1,0 +1,97 @@
+import json
+from typing import List
+import pickle
+import pandas as pd
+from presidio_evaluator.experiment_tracking.experiment_tracker import ExperimentTracker
+from presidio_evaluator.evaluation.model_error import ModelError
+from pathlib import Path
+
+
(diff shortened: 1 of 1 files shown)


presidio_evaluator/experiment_tracking/local_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+                         "experiment_params.json")
+        print(f"saving experiment data to directory {self.experiment_name}")
+        with file_path.open('w') as json_file:
+            json.dump(self.__dict__, json_file)
+
+
+if __name__ == "__main__":
+
+    directory = Path(Path.cwd(), 'experiments')
+    experiment_name = "my_first_experiment"
```
tranguyen221, line 84: The code looks great to me. Just a minor comment. Should we add the datetime in the experiment name, so the user can track when the experiment is done?

tranguyen221 (commented):
The code looks great to me. Just add a minor comment.

presidio_evaluator/experiment_tracking/local_tracker.py
melmatlis: It's a great idea, I will add it as a parameter in the `.json` file, this way the experiment_name will remain clear and easy to remember but the the date time of the experiment execution will be stored. What do you think?

tranguyen221: I think that would be a great option. Just a question, if a user runs the same experiment name but in 2 different timestamps, the old results will be overwritten by the new one?

melmatlis: At the moment yes.

tranguyen221: Ok. In this case, I think we add datetime in the .json file.

melmatlis: done

tranguyen221: Great! Thank Mel!

tranguyen221 (approved):

PR #63, opened 2022-12-19
Tranguyen/improve token evaluation

tranguyen221:
The code generates the metrics at the span level:
1. Add predict_span function to PresidioAnalyzerWrapper class to return the predicted outcomes in Span format
2. Implement SpanEvaluator class to evaluate the outcomes at the span level
3. Implement SpanOutput to hold the model prediction output for analysis purposes
4. Implement SpanEvaluationResult to hold the final evaluation results
5. Add 4_Evaluate_Presidio_Analyzer_Span_Level.ipynb for an example of evaluating presidio at the span level
6. Update README.md

Next steps:
- Implement F1 and F2 score for strict and flexible evaluation
- Add unit tests for the additional classes
- Modify the local experiment tracking to log the errors and results of span evaluation
- Add the notebooks for evaluating other models (spacy, text analytics, etc.) at the span level


presidio_evaluator/evaluation/span_output.py
```diff
@@ -0,0 +1,74 @@
...
+            f"Output type: {self.output_type}\n"
+            f"Overlap score: {self.overlap_score}\n"
+            f"Full text: {self.full_text}\n"
+            f"Gold span: {self.gold_span}\n"
+            f"Predicted span: {self.pred_span}\n"
+        )
+
+
+    @staticmethod
+    def get_spans_output_by_type(output_type:str, outputs=List["SpanOutput"], entity=None):
```
melmatlis, line 37:
```suggestion
    def get_spans_output_by_type(output_type:str, outputs=List["SpanOutput"], entity: Optional[Union[List, str]] =None) -> List["SpanOutput"]:
```

presidio_evaluator/models/presidio_analyzer_wrapper.py
```diff
@@ -65,6 +66,30 @@ def predict(self, sample: InputSample) -> List[str]:
...
+    def predict_span(self, sample: InputSample) -> List[Span]:
+        """
+         Returns the predicted spans from the evaluated model
+        :param sample: Sample to be evaluated
+        :return: List of span in Span object
+        """
+        results = self.analyzer_engine.analyze(
+            text=sample.full_text,
+            entities=self.entities,
+            language="en",
```
melmatlis, line 78: There is another PR which made the language an external dependency. Perhaps we should conform and add language as an argument

notebooks/4_Evaluate_Presidio_Analyzer.ipynb
```diff
@@ -5,23 +5,39 @@
    "id": "847acd88",
    "metadata": {},
    "source": [
-    "Evaluate Presidio Analyzer using the Presidio Evaluator framework"
+    "# Evaluate Presidio Analyzer using the Presidio Evaluator framework\n",
```
melmatlis, line 8: The output of the notebook is very large, may you please clear the outputs of the notebook and recommit it please?
Thank you

melmatlis (changes requested):
I'm so excited to have this capability in the presidio research.

tests/mocks/model_mock.py
```diff
@@ -57,3 +57,4 @@ def predict(self, sample: InputSample) -> List[str]:
             return sample.tags
         else:
             return ["O" for i in range(len(sample.tags))]
+
```
omri374, line 60: Once the new structure is in place, we would need to support it with unit tests (especially for the evaluation logic)

omri374 (changes requested):
I'm positing my initial comments here, but Trang and I also did some brainstorm on potentially making evaluation work always on both span and token levels, and have the user decide which metrics he/she are interested in (token level based, span based or custom). 

For that, we discussed the option of having the model returning both predicted tokens and predicted spans. Since each model supports only one initially (e.g. transformers would return spans, but CRF would return tokens), we can add a helper function that translates `span_to_tag` or `tag_to_span`. Then, the evaluator would calculate metrics on both outputs.

Additionally, we discussed the potential de-coupling of the model class and evaluator class, in a way that would allow the evaluator to accept a model output and run evaluation on it, instead of having to call the `predict` method of a model (which allows users to use the evaluator without having to have a model which conforms with the presidio-evaluator API). 

Trang will do some prototyping and propose a design. This could end up being non-backward-compatible but we'll discuss this and weigh pros and cons.

For example, a potential model output class:
```python
class ModelPrediction:
    def __init__(
        self, predicted_tags: Optional[List[str]]=None, predicted_spans: Optional[List[Span]]=None
    ):
        """
        Hold the results of a model prediction on a token and/or span level
        :param predicted_tags: List of predicted tags (annotation per token)
        :param predicted_spans: List of predicted spans (annotation per span)
        """
        self.predicted_tags = predicted_tags
        self.predicted_spans = predicted_spans

```

Issue #64, opened 2022-12-20
DataGenerator - templates containing square brackets treated as entities to be replaced


melmatlis:
Templates which contain square brackets fail on generating new fake data. Square brackets are converted to curly brackets to be treated as entities i.e -> "{{entity}}". Method `_prep_templates` responsible for the replacement.
Is this the desired behavior?

How to reproduce the error:
```

from faker import Faker
fake = RecordsFaker(fake_data_df, local="en_US")
data_generator = PresidioDataGenerator(
    custom_faker=fake, lower_case_ratio=0
)

provider_list = [HospitalProvider, IpAddressProvider,NationalityProvider,  AgeProvider, AddressProviderNew, PhoneNumberProviderNew, OrganizationProvider]
for provider in provider_list:
    fake.add_provider(provider)

template = "My name is {{name}} and I [travel] to the factory"


fake_records = data_generator.generate_fake_data(
    templates=[template], n_samples=1
)
list(fake_records)
```
Returns:

>AttributeError: Failed to generate fake data based on template "My name is {{name}} and I {{travel}} to the factory".You might ?need to add a new Faker provider! Unknown formatter 'travel'

omri374:
It is by design (as the previous version used square brackets and we wanted to have some backward compatibility, but I don't see an issue with removing this, especially if it causes issues with generated templates.

omri374:
Reference in code: https://github.com/microsoft/presidio-research/blob/d34e4766623457dd508638b1a466a73031abe350/presidio_evaluator/data_generator/presidio_data_generator.py#L139

melmatlis:
Solved by #65

PR #65, opened 2022-12-26
Remove option to place faker entities inside square brackets

melmatlis:
Enable the option for templates to contain information in square brackets without being parsed as entities.

Response to issue : #64


omri374 (approved):
Thanks!

PR #71, opened 2023-01-30
Refactor evaluation to allow span based metrics

tranguyen221:
- Implement the compare_span function in Evaluator class which takes lists of annotated_span and predicted_span and then generates the evaluation output in numbers
- Implement the get_overlap_ratio function in Span class to calculate the overlapping ratio between annotated and predicted offsets.
- Add the unittest for those functions


melmatlis:
A general comment regarding the `data_object.py` file and unrelated to the PR.
The file is really long and contains 2 classes, perhaps we can add a **separate** task to add a folder for "data objects" and split the classes to separate files?
@omri374 @tranguyen221  what are your thoughts?

omri374:
@melmatlis I agree, but suggest to wait with this until we finalize all the changes, in order not to make unnecessary conflicts.

omri374 (commented):
Publishing initial comments as discussed in our code review

presidio_evaluator/evaluation_helpers.py
```diff
@@ -0,0 +1,156 @@
...
+    pred_set = set(pred_range)
+
+    overlaps = true_set.intersection(pred_set)
+
+    return overlaps
+
+def span_compute_actual_possible(results: dict) -> dict:
+    """
+    Takes a result dict that has been output by compute metrics.
+    Returns the results dict with actual, possible populated.
```
melmatlis, line 46: I would propose to update the doc string further and to explain what's "Actual"  and "possible" refer to.
Add the formulas into the docstring as well

presidio_evaluator/evaluation_helpers.py
```diff
@@ -0,0 +1,156 @@
...
+def span_compute_precision_recall(results: dict, partial_or_type) -> dict:
+    """
+    Takes a result dict that has been output by compute metrics.
+    Returns the results dict with precison and recall populated.
+    When the results dicts is from partial or ent_type metrics, then
+    partial_or_type=True to ensure the right calculation is used for
+    calculating precision and recall.
+    """
+
+    actual = results["actual"]
```
melmatlis, line 81: Based on the schema you created, we should have a class named EvaluationResult. Are the results of type dict as an intermediate solution?

melmatlis (changes requested):
Hey Trang, thanks for the PR. A lot of work went into the complicated modifications. Thank you for these.
May we please have a peer review session on the `evaluator.py` file. I need some guidance on the code readability. Thank you

omri374:
Implemented in https://github.com/microsoft/presidio-research/pull/155

PR #73, opened 2023-02-07
Tranguyen/new eval prototype

tranguyen221:
@omri374 as discussed in our code review, I created a new folder presidio_evaluator_2 which contains the prototype of all classes needed for the evaluator refactoring.


omri374 (approved):
Thanks!

A few suggestions going forward:
1. Let's consider deleting the other files and tests so that we could always run unit tests on the newly added code
2. Docstring format should be unified. In Presidio we use sphinx/restructuredtext but here there's a mix of two formats (see https://betterprogramming.pub/3-different-docstring-formats-for-python-d27be81e0d68)
3. For classes we have two options for docstrings: either put everything (description + init arguments) on the class docstring, or put everything on the init docstring. No need to have arguments defined in both.

Should we consider validating for flake8? WDYT?

PR #74, opened 2023-03-02
added a new method for tag_to_span

omri374:
Helper method to perform tag labeling to span labeling (opposite of `span_to_tag` for cases where models return tags but we want to calculate span based metrics.

### DIFF
--- a/presidio_evaluator/tag_to_span.py
+++ b/presidio_evaluator/tag_to_span.py
@@ -1,0 +1,62 @@
+from typing import List, Optional
+from spacy.tokens import Doc
+
+from presidio_evaluator import Span
+
+
+def _to_io(tag):
+    if "-" in tag:
+        return tag[2:]
+    return tag
+
+
+def tag_to_span(tokens: Doc, tags: List[str]) -> List[Span]:
+    """
+    Transforms a list of tags to a list of spans, while ignoring labeling scheme.
+    :param tokens: List of spaCy tokens
+    :param tags: List of tags (e.g. ["O", "B-PER", "I-PER", "U-LOC"])
+    :return: List of spans
+    """
+
+    # If IO, translate to BIO for easier manipulation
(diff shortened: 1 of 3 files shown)


presidio_evaluator/tag_to_span.py
```diff
@@ -0,0 +1,62 @@
...
+
+def tag_to_span(tokens: Doc, tags: List[str]) -> List[Span]:
+    """
+    Transforms a list of tags to a list of spans, while ignoring labeling scheme.
+    :param tokens: List of spaCy tokens
+    :param tags: List of tags (e.g. ["O", "B-PER", "I-PER", "U-LOC"])
+    :return: List of spans
+    """
+
+    # If IO, translate to BIO for easier manipulation
```
tranguyen221, line 21: should it be "if it is not IO, translate to IO for easier manipulation"?

tests/test_tag_to_span.py
```diff
@@ -0,0 +1,46 @@
...
+            [[2, 3, "name"], [3, 4, "loc"]],
+        ),
+        (
+            "Dan is Washington",
+            ["U-name", "O", "U-loc"],
+            [[0, 1, "name"], [2, 3, "loc"]],
+        ),
+    ],
+)
+def test_tag_to_span(spacy_en_core_web_sm, text, tags, ents):
```
tranguyen221, line 34: It's great! The function works well. I also add the other test cases and all of them passed. Thanks @omri374 for working on it.

tranguyen221 (commented):
Thanks @omri374. I also tested the function with the other cases and it works well!

presidio_evaluator/tag_to_span.py
omri374: Yes! good catch

tests/test_tag_to_span.py
omri374: Thanks!

tranguyen221 (approved):

PR #120, opened 2025-01-31
Extension of the Plotter class with the possibility of saving multipl…

mar-fij:
Extension of the Plotter class with the possibility of saving multiple plots to the output folder.


omri374:
Fixed in https://github.com/microsoft/presidio-research/pull/155

Issue #127, opened 2025-02-18
Passing a PresidioAnalyzer to the Evaluator class overrides configuration


omri374:
When passing an `AnalyzerEngine` class to the `Evaluator` class for evaluation, the `AnalyzerEngine` is wrapped with a default `PresidioAnalyzerWrapper`. This wrapper has some default values that are overriding the configuration given through the presidio analyzer.

To reproduce:
```py
from presidio_evaluator import InputSample

from presidio_evaluator.evaluation import Evaluator
from presidio_analyzer import AnalyzerEngine

analyzer = AnalyzerEngine(default_score_threshold=1.0)
evaluator = Evaluator(analyzer)
evaluator.evaluate_all(dataset)

```
Should filter out all predictions with score < 1.0, but in reality it filters only with 0.4 with is the default value in `PresidioAnalyzerWrapper`, specifically here:
https://github.com/microsoft/presidio-research/blob/9b0aaff5920e6974d22cae7980bfe58b7fcd60eb/presidio_evaluator/models/presidio_analyzer_wrapper.py#L65

PR #133, opened 2025-03-02
fix: support non-en usage of evaluator with analyzer_engine

tianjianjiang:
Hi,

This simple PR may or may not actually be a breaking change due to the newly added default kwarg of language. It is required, however, by non-en usages. 🙏


omri374:
While this fixes the issue, it is specific to the use of the analyzer engine as the input to the Evaluator. We are aware of this issue (#127), the best way (currently) to avoid this is to use the `PresidioAnalyzerWrapper` directly:
```py
wrapper = PresidioAnalyzerWrapper(analyzer, ...) # analyzer with additional runtime args
evaluator = Evaluator(wrapper)
```

PR #157, opened 2025-07-20
Span eval docs and more

omri374:
Created evaluation docs including the comparison between token and span, and between presidio-evaluator and other approaches (semeval 13, CONNL)

Several enhancements and bug fixes arising from the work on the documentation

### DIFF
--- a/docs/span_evaluation.md
+++ b/docs/span_evaluation.md
@@ -1,0 +1,105 @@
+# Span Evaluation in Presidio Evaluator
+
+This document explains the span evaluation process implemented in Presidio Evaluator, covering how spans are created,
+matched, and evaluated, along with comparisons to other evaluation paradigms.
+
+## Span Creation and Processing
+
+### Span Creation
+
+Spans are created from token-level annotations in the input data. Each span represents a continuous sequence of tokens
+with the same entity type annotation. The basic properties of a span include:
+
+- `entity_type`: The type of entity (e.g., PERSON, LOCATION)
+- `entity_value`: The actual text of the entity
+- `start_position` and `end_position`: Character-level boundaries
+
+
+#### Span Normalization
+
+For more advanced processing, spans also include normalized versions of the text:
+
+- `normalized_tokens`: List of normalized tokens that make up the span (typically lowercased and with special characters
+  removed)
+- `normalized_start_index` and `normalized_end_index`: Character indices in the normalized text
+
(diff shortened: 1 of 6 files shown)


docs/span_evaluation.md
```diff
@@ -0,0 +1,105 @@
...
+> For a detailed breakdown of different matching scenarios and examples, see
+> the [Span Matching Strategies](span_matching_strategies.md) document.
+
+## Metric Calculation
+
+The evaluator calculates both per-entity-type metrics and global PII metrics:
+
+### Per-Entity-Type Metrics
+
+- **Precision**: TP / (TP + FP)
```
negruber1, line 80: We currently calculate per type precision and recall as precision=TP/(total predicted per type) and recall=TP/(total annotated per type)

docs/span_matching_strategies.md
```diff
@@ -0,0 +1,131 @@
...
+    - Text: "John Smith Johnson"
+    - Annotation: [PERSON, PERSON, PERSON]
+    - Prediction: [PERSON, LOCATION, PERSON]
+    - PERSON IoU = 0.67, LOCATION IoU = 0.33
+    - If threshold = 0.5: PERSON is a match but wrong type for LOCATION portion
+    - Result: Both type mismatch and partial match errors
+
+## Real-world Examples
+
+### Example 1: Complex Name with Multiple Parts
```
negruber1, line 99: Isn't the FN only for TITLE? because we matched the PERSON spans (assuming IoU>threshold)?

omri374: Good catch

docs/span_evaluation.md
negruber1: Maybe we should only state that we do not use the "typical" precision/recall formula

docs/span_evaluation.md
```diff
@@ -0,0 +1,105 @@
...
+Union (IoU) approach. This can be either character-based or token-based, controlled by the `char_based` parameter.
+
+### IoU Calculation
+
+- **Character-based IoU**: Calculates the character overlap between spans
+- **Token-based IoU**: Calculates the token overlap between spans
+
+IoU = (Intersection) / (Union)
+
+An IoU threshold (default: 0.75) determines whether spans match sufficiently.
```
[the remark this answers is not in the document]
omri374: Yes! Good catch

PR #162, opened 2025-12-02
Fix swapped annotation/prediction fields in _get_model_error

rniu-scai:
Hi, I noticed that `ModelError.get_fps_dataframe` and `ModelError.get_fns_dataframe` had their "annotated" and "prediction" columns incorrectly swapped, therefore the explanation would not match what was in these columns.

This is a simple bug fix that resolves the root of this issue.


omri374:
Thanks!

omri374 (approved):

PR #169, opened 2026-03-26
Presidio evaluator redesign towards v0.3

omri374:
This PR replaces the tightly-coupled evaluate_all() monolith with a clean three-layer pipeline: 
```Model → CanonicalMapper → Evaluator```. 

Main changes:
- The evaluator no longer owns model inference or entity mapping. Each step is an independent function call with a shared DataFrame as the interface. `CanonicalMapper` (new entity_mapping package) performs a four-tier auto-resolution (EXACT → COUNTRY → FUZZY → PENDING) that surfaces unknown labels explicitly instead of silently dropping them. 
- Non-Presidio model wrappers (FlairModel, SpacyModel, StanzaModel, AzureAITextAnalyticsWrapper) are removed; models should be added via Presidio's recognizer framework. 
- Tooling is modernized: Poetry → uv, ruff added, pre-commit hooks, Python minimum raised to 3.11.

Note to reviewers: There are 121 files changed. Please don't read them all. Start with the two ADRs (ADR-001, ADR-002) for design intent, then migration-guide.md cover all the new contracts. Run tests with `uv run pytest -m "not integration"` and review the notebooks (4,5,6) too.

What (almost) didn't change:
- data_objects
- data generator
- notebooks 1,2,3
- dataset formatters
- experiment tracking

### CODE REVIEW

File: presidio_evaluator/models/presidio_analyzer_wrapper.py [hunk 1]
```diff
@@ -14,17 +12,17 @@
 class PresidioAnalyzerWrapper(BaseModel):
     def __init__(
         self,
-        analyzer_engine: Optional[AnalyzerEngine] = None,
-        entities_to_keep: List[str] = None,
+        analyzer_engine: AnalyzerEngine | None = None,
+        entities_to_keep: list[str] | None = None,
```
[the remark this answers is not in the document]
omri374: Fixed

docs/adr/ADR-001-simplified-evaluation-pipeline.md
```diff
@@ -0,0 +1,160 @@
...
+- **`calculate_score_on_df` already exists and is tested for spans** — `SpanEvaluator.calculate_score_on_df` is already implemented and covered by tests in `test_span_evaluator.py`. A similar API can be added for `TokenEvaluator` as part of this ADR.
+- **Eliminates the `EvaluationResult` per-sample carrier** — removes a class whose only purpose is to ferry data between `evaluate_all()` and `get_results_dataframe()`.
+
+### Negative / Trade-offs
+
+- **`predict_dataset` materializes all predictions in memory** — for very large datasets, the full DataFrame is held in RAM. The current `evaluate_all()` loop can in principle be streamed (though it isn't today).
+- **`evaluate_all()` becomes a thin wrapper** — code that currently calls `evaluate_all()` and inspects `List[EvaluationResult]` directly will need to be updated if it relies on the per-sample `EvaluationResult` structure. Backward compatibility is preserved at the call site, but the internal representation changes.
+- **Schema enforcement is implicit** — the 5-column schema is a convention, not enforced by a type. Callers that hand-construct DataFrames must respect column names and types.
```
negruber1, line 125: Should we validate the dataframe schema?

presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -0,0 +1,884 @@
...
+
+        # BIO-stripped version of each label (original label remains the dict key)
+        self._stripped: dict[str, str] = {
+            label: EntityHierarchy._strip_bio(label) for label in self._labels
+        }
+
+        self._records: dict[str, _Resolution] = {}
+        self._auto_resolve()
+
+    def get_mapped_results_dataframe(
```
negruber1, line 397: If the user calls get_mapped_results_dataframe after defining some custom entities, wouldn't this wipe out these user defined entities? For example, if we get warning regarding the different hierarchy levels, and get_mapped_results_dataframe is called again to resolve the conflict, self._records.clear() will remove all existing mappings including the manually defined ones

presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -0,0 +1,884 @@
...
+    # ── State ────────────────────────────────────────────────────────────────
+
+    @property
+    def pending(self) -> list[str]:
+        """Labels not yet resolved, in alphabetical order."""
+        return sorted(label for label in self._labels if label not in self._records)
+
+    # ── Mutation ─────────────────────────────────────────────────────────────
+
+    def map(self, mappings: dict[str, str | None]) -> CanonicalMapper:
```
negruber1, line 460: Should manual entities be exported for future use/reproducibility? And also, should we make sure the manual entities are consistent across models?

omri374:
@negruber1 thanks! great points. I'll release a new ADR soon with the things we mentioned yesterday, and after we agree on this will update the code including with your suggestions here.

omri374:
@negruber1 something I thought about: If we look at all the collisions between annotated and predicted entities to surface what needs to be fixed, are we adding a confounding effect to the comparison in any way? Could be minor/negligible though.

negruber1:
@omri374 I think that if the conflicts are resolved by the user BEFORE seeing evaluation results, the confounding effect is probably minor: If there is a "discovery trial" where you get to see the model's vocabulary without calculating the evaluation metrics, the user decides to resolve based on the existing labels and not which labels the model gets wrong (We can even recommend that on the mapper README). We can always evaluate at all hierarchies and make this available for the user regardless of what mappings they decided to define manually. 
I think that because we include human input, we can't avoid a confounding effect entirely.

omri374:
> @omri374 I think that if the conflicts are resolved by the user BEFORE seeing evaluation results, the confounding effect is probably minor: If there is a "discovery trial" where you get to see the model's vocabulary without calculating the evaluation metrics, the user decides to resolve based on the existing labels and not which labels the model gets wrong (We can even recommend that on the mapper README). We can always evaluate at all hierarchies and make this available for the user regardless of what mappings they decided to define manually. I think that because we include human input, we can't avoid a confounding effect entirely.

Thanks, completely agree.

omri374:
closing this PR in favor of an updated implementation

PR #182, opened 2026-07-28
Fix token-based span IoU to be position-aware

omri374:
## Problem

`SpanEvaluator.calculate_iou` with `char_based=False` (and the token path of `_calculate_combined_iou`) compared **sets of token strings**, discarding token positions. Two failure modes:

1. **Duplicate words**: in `"Michael met Michael"`, annotating the first *Michael* while predicting the second gave IoU = 1.0 and was counted as a true positive.
2. **Shared words across different entities**: gold `"Human Rights"` vs prediction `"Civil Rights"` produced IoU = 1/3 through the shared token *rights*, despite the spans being disjoint.

## Fix

- `Span` gains an optional `normalized_start_indices` field (per-token start indices of the normalized tokens), populated by `SpanEvaluator._create_spans` and preserved by `_merge_adjacent_spans`.
- Token-level IoU (both `calculate_iou` and `_calculate_combined_iou`) now compares sets of `(start_index, token)` pairs, making it position-aware and duplicate-safe.
- Spans constructed without per-token indices (e.g. manually built ones) fall back to the previous string comparison, but guarded by a positional-overlap check so disjoint spans always get IoU = 0.

## Tests (TDD)

Added tests written first to reproduce the bug, then made green:

- `test_token_iou_same_word_different_position_is_zero` — "Michael met Michael" end-to-end: FN + FP, not TP.
- `test_token_iou_shared_word_entities_do_not_overlap` — "Human Rights" vs "Civil Rights" IoU = 0.
- `test_token_iou_duplicate_tokens_within_span` — "Michael Michael" vs first "Michael" gives IoU 0.5, not 1.0.
- `test_token_iou_exact_match_still_one` — sanity: exact match still IoU 1.0 / TP.
- `test_combined_token_iou_duplicate_tokens` — combined IoU doesn't collapse duplicates.
- `test_token_iou_manual_spans_disjoint_positions` — fallback path for spans without per-token indices.

Full suite: 669 passed (integration tests excluded).

### DIFF
--- a/presidio_evaluator/evaluation/span_evaluator.py
+++ b/presidio_evaluator/evaluation/span_evaluator.py
@@ -228,7 +228,7 @@ class SpanEvaluator(BaseEvaluator):
             carry per-token start indices
         """
         tokens = span.normalized_tokens or []
-        indices = getattr(span, "normalized_start_indices", None)
+        indices = span.normalized_start_indices
         if indices is None or len(indices) != len(tokens):
             return None
         return set(zip(indices, tokens, strict=True))
@@ -599,7 +599,7 @@ class SpanEvaluator(BaseEvaluator):
                                 token_start=current_token_start,
                                 current_tokens=current_tokens,
                                 idx=idx,
-                                normalized_satrt_indices=normalized_start_indices,
+                                normalized_start_indices=normalized_start_indices,
                                 normalized_tokens=normalized_tokens,
                             ),
                         )
@@ -625,7 +625,7 @@ class SpanEvaluator(BaseEvaluator):
                                 token_start=current_token_start,
                                 current_tokens=current_tokens,
                                 idx=idx,
-                                normalized_satrt_indices=normalized_start_indices,
+                                normalized_start_indices=normalized_start_indices,
                                 normalized_tokens=normalized_tokens,
                             ),
                         )
@@ -654,7 +654,7 @@ class SpanEvaluator(BaseEvaluator):
                         token_start=current_token_start,
                         current_tokens=current_tokens,
                         idx=df.index[-1] + 1,
-                        normalized_satrt_indices=normalized_start_indices,
+                        normalized_start_indices=normalized_start_indices,
                         normalized_tokens=normalized_tokens,
                     ),
                 )
@@ -667,7 +667,7 @@ class SpanEvaluator(BaseEvaluator):
         token_start: int,
         current_tokens: list[str],
         idx: int,
-        normalized_satrt_indices: list[int],
+        normalized_start_indices: list[int],
         normalized_tokens: list[str],
     ):
         return Span(
@@ -676,12 +676,12 @@ class SpanEvaluator(BaseEvaluator):
             start_position=start_indices[0],
             end_position=start_indices[-1] + len(current_tokens[-1]),
             normalized_tokens=normalized_tokens,
-            normalized_start_index=min(normalized_satrt_indices),
+            normalized_start_index=min(normalized_start_indices),
             normalized_end_index=self._get_normalized_end_index(
                 normalized_tokens,
-                normalized_satrt_indices,
+                normalized_start_indices,
             ),
-            normalized_start_indices=normalized_satrt_indices,
+            normalized_start_indices=normalized_start_indices,
             token_start=token_start,
             token_end=idx,
         )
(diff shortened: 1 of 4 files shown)


presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -188,6 +214,25 @@ def calculate_iou(
...
+        Comparing tokens together with their positions prevents identical words at
+        different positions (e.g. both occurrences in "Michael met Michael") from
+        being treated as the same token.
+
+        :param span: Span to extract positional tokens from
+        :return: Set of (start index, token) pairs, or None if the span does not
+            carry per-token start indices
+        """
+        tokens = span.normalized_tokens or []
+        indices = getattr(span, "normalized_start_indices", None)
```
omri374: Why getattr and not direct?

presidio_evaluator/evaluation/span_evaluator.py
negruber1, line 1308: > **Single-source the position-aware-vs-fallback policy.** It seems that the same *policy* is implemented independently in both `calculate_iou` and `_calculate_combined_iou`:
>
> 1. build positional tokens via `_positional_tokens(...)` for every span,
> 2. check whether all spans carry indices,
> 3. if not, guard the string-set fallback with a positional-overlap check (`intersect(...)`) so identical tokens at disjoint positions can't match,
> 4. assemble `set(normalized_tokens)` fallbacks,
> 5. compute `intersection/union`.
>
> The bug-critical step is (3) — and it's already encoded *differently* in the two places: `calculate_iou` short-circuits the whole comparison with a single `return 0.0`, while `_calculate_combined_iou` applies the guard per-prediction inside the loop, excluding non-overlapping predictions from `pred_tokens`. Those are equivalent today, but because the guard has two independent encodings, a future change to one won't mechanically propagate to the other — and we'd end up with two subtly different definitions of "token IoU" depending on whether matching went through the single- or combined-match path. That's a hard bug to notice since both still return plausible numbers. 

Should we create a helper that is used in both functions?

tests/evaluation/test_span_evaluator.py
```diff
@@ -1535,3 +1535,162 @@ def test_model_error_start_end_positions(
         assert error.end == expected["end"], (
             f"In {scenario}, {expected['error_type']} end expected {expected['end']}, got {error.end}"
         )
+
```
RonShakutai, line 1538: The tests cover the negative cases well, would it be worth adding one positive combined case ?

presidio_evaluator/evaluation/span_evaluator.py
omri374: Great suggestion, fixed

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -178,15 +189,75 @@ def calculate_iou(
...
+            SpanEvaluator._positional_tokens(pred_span) for pred_span in pred_spans
+        ]
+
+        if ann_tokens is not None and all(
+            tokens is not None for tokens in pred_token_sets
+        ):
+            pred_tokens: set = set()
+            for tokens in pred_token_sets:
+                pred_tokens.update(tokens or set())
+        else:
```
RonShakutai, line 232: in the fallback once two spans overlap we compare plain words again, so a repeated/shared word can still cause a wrong match, is this intentional ? or hsould stay position aware too ?

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -178,15 +189,75 @@ def calculate_iou(
...
+        different positions (e.g. both occurrences in "Michael met Michael") from
+        being treated as the same token.
+
+        :param span: Span to extract positional tokens from
+        :return: Set of (start index, token) pairs, or None if the span does not
+            carry per-token start indices
+        """
+        tokens = span.normalized_tokens or []
+        indices = span.normalized_start_indices
+        if indices is None or len(indices) != len(tokens):
```
RonShakutai: if the token list and index list have different lengths, it returns quietly None, 
Should we raise warning / exception ?

presidio_evaluator/evaluation/span_evaluator.py
omri374: This fallback is here for an edge case where someone created spans manually and they don't have the token indices. I was thinking if this fallback is even necessary because it's highly unlikely. In any case, you are right, but there's no way out of this (again, in the case where we don't have something that we calculate ourselves)

presidio_evaluator/evaluation/span_evaluator.py
omri374: Good point

tests/evaluation/test_span_evaluator.py
omri374: Good point. There are some positive test cases in the existing test suite, but I'll harden it.

PR #184, opened 2026-07-29
Fix index-sensitive span merging in SpanEvaluator

omri374:
## Summary

`SpanEvaluator` scoring results depended on the DataFrame index of `results_df` — the native output of `predict_dataset()` (global RangeIndex) silently produced wrong span counts for every sentence except the first.

**Root cause:** `_create_spans` records `token_start`/`token_end` as positions *within the sentence* (via `enumerate`), but `_are_spans_adjacent` sliced the sentence DataFrame with label-based `df.loc[token_end : token_start - 1]`. For any sentence whose first row label isn't 0, that lookup reads the wrong rows — or returns nothing, and an empty between-tokens list counts as "adjacent". Result: same-type spans separated by regular words were silently merged (the two PERSON spans in *"John visited Berlin with Mary"* became one), deflating `num_annotated` / `num_predicted` / `true_positives`. Because gold and predictions were distorted symmetrically, headline precision/recall could still look plausible.

## Changes

- `_are_spans_adjacent`: positional `.iloc` slicing instead of label-based `.loc`.
- `_create_spans`: final span's end position is `len(df)` instead of `df.index[-1] + 1`; space-accounting guard compares positions (`idx > 0`) instead of index labels.
- Regression test `test_multi_sentence_df_does_not_merge_separated_same_type_spans`: two identical sentences must score identically regardless of where they sit in the dataset (fails before this fix with 3 spans instead of 4).
- CHANGELOG entry + version bump to 0.3.2 for the upcoming release.

Scoring is now index-agnostic: global, per-sentence, or any other index produces identical results.

## Impact

Any prior multi-sentence evaluation run through `predict_dataset()` was affected; re-running will shift reported numbers (that's the fix, not a regression). Existing tests never caught this because they all use single-sentence DataFrames whose labels coincide with positions.

## Test plan

- [x] New regression test fails on `main`, passes here
- [x] Full unit suite: 672 passed

### DIFF
--- a/presidio_evaluator/evaluation/span_evaluator.py
+++ b/presidio_evaluator/evaluation/span_evaluator.py
@@ -157,9 +157,11 @@ class SpanEvaluator(BaseEvaluator):
         # token_start/token_end are positions within the sentence, so slice
         # positionally — the DataFrame's index labels are caller-defined
         # (e.g. global across sentences) and must not be used as positions.
-        between_tokens = (
-            df["token"].iloc[span1.token_end : span2.token_start or 0].tolist()
-        )
+        if span1.token_end is None or span2.token_start is None:
+            raise ValueError(
+                "Spans must have token_start/token_end set to check adjacency",
+            )
+        between_tokens = df["token"].iloc[span1.token_end : span2.token_start].tolist()
         non_skip_tokens = [
             tok for tok in between_tokens if tok.lower().strip() not in self.skip_words
         ]
@@ -613,19 +615,11 @@ class SpanEvaluator(BaseEvaluator):
         current_tokens = []
         current_start_indices = []
         current_token_start: int = 0
-        curr_char_position = 0
-        token_position = 0  # Add token position counter
 
         for idx, (_, row) in enumerate(df.iterrows()):
             entity_type = row[column]
             token = row["token"]
             token_start = row["start_indices"]
-            token_length = len(token)
-            # If this isn't the first token, add space before it
-            if idx > 0:
-                curr_char_position += 1  # Account for space between tokens
-
-            token_end = curr_char_position + token_length
 
             if entity_type == "O":
                 if current_entity_type and current_tokens:
@@ -649,8 +643,6 @@ class SpanEvaluator(BaseEvaluator):
                     current_start_indices = []
                     current_token_start = 0
 
-                curr_char_position = token_end
-                token_position += 1  # Increment token position
                 continue
 
             if entity_type != current_entity_type:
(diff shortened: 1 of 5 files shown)


presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -154,11 +154,12 @@ def _are_spans_adjacent(self, span1: Span, span2: Span, df: pd.DataFrame) -> boo
...
-        # Slice tokens between span1 and span2 using the row indices
-        between_tokens = df.loc[
-            span1.token_end : (span2.token_start or 0) - 1,
-            "token",
-        ].tolist()
+        # token_start/token_end are positions within the sentence, so slice
+        # positionally — the DataFrame's index labels are caller-defined
+        # (e.g. global across sentences) and must not be used as positions.
+        between_tokens = (
+            df["token"].iloc[span1.token_end : span2.token_start or 0].tolist()
```
negruber1: If `token_start` is `None`, `or 0` turns it into `0`, the slice comes back empty, and empty means "adjacent" , so the two spans merge. That's the same empty-slice-means-merge failure the PR just fixed, reached by a different route. (`token_end is None` fails the other way: pandas reads it as an open slice, scans from token 0, and wrongly returns "not adjacent".)

Both fields default to `None` on `Span`, and only `_create_spans` fills them in. Nothing breaks today because token_start=None is not possible, but it could be misleading because the "or 0" statement can't distinguish "token 0" from "unknown" and this might create issues in future code changes.

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -621,7 +622,7 @@ def _create_spans(self, df: pd.DataFrame, column: str) -> list[Span]:
             token_start = row["start_indices"]
             token_length = len(token)
             # If this isn't the first token, add space before it
-            if idx > df.index[0]:
+            if idx > 0:
```
negruber1: I think this if block is dead code. `curr_char_position`, `token_end`, `token_position`, and `token_length` only feed each other, nothing they compute ever reaches a `Span`, since `__create_span` derives all offsets from the `start_indices` column.

omri374: Confirmed — `curr_char_position`, `token_end`, `token_position`, and `token_length` form a closed cycle that never reaches a `Span`; all character offsets come from the `start_indices` column. Removed the whole group (including the `idx > 0` space-accounting block) in dc1308b. Full suite still passes (705 passed).

presidio_evaluator/evaluation/span_evaluator.py
omri374: Good catch — fixed in dc1308b. Dropped the `or 0` and `_are_spans_adjacent` now raises `ValueError` if either `token_start`/`token_end` is `None`, so an unset position fails loudly instead of silently reading as "adjacent" (or "not adjacent" for `token_end=None`).

negruber1 (approved):

PR #186, opened 2026-07-30
feat: branch-level aliases (LOC->LOCATION, ORG->ORGANIZATION)

RonShakutai:
## Problem

Leaf nodes in the hierarchy carry alias lists (`"EMAIL_ADDRESS": ["EMAIL", ...]`), but branch (non-leaf) nodes are dicts with nowhere to declare synonyms. So coarse dataset labels like TAB's `LOC`/`ORG` existed as separate empty leaves and could only ever match a model's `LOCATION`/`ORGANIZATION` at the **branch** level, never **exact** — understating exact-level scores for every model on such a corpus.

## Fix

Add a reserved `_aliases` key so a branch can declare raw aliases:

```python
"LOCATION": {"_aliases": ["LOC"], "ADDRESS": {...}, ...}
```

Every tree-walk skips the key, so it never becomes a canonical entity. `LOC`/`ORG` move onto the `LOCATION`/`ORGANIZATION` branches.

Once they stop being canonical entities, the canonical-name lookups have to learn about aliases:

- `to_branch()` / `get_depth()` now resolve aliases first. Without this `to_branch("LOC")` would silently return `"LOC"` instead of `"LOCATION"` — projecting into a *different branch* rather than failing. This also makes existing leaf aliases (`COMPANYNAME`, `QQ`) work for the first time.
- `add_alias()` accepts an alias as its subject, so `add_alias("LOC", ...)` works.
- `_find_node()` no longer matches the reserved key, which let `add_alias("_aliases", x)` attach `x` to whichever branch the walk hit first.
- `add_alias()` raises `ValueError` instead of silently no-opping when a descendant already claims the name (branch aliases are applied before the descent into their own subtree, so the descendant wins). The hierarchy is rolled back.

## Breaking

`LOC`/`ORG` are no longer canonical entities:

- `canonicalize("LOC") == "LOCATION"` (was `"LOC"`); absent from `all_canonical_entities` and `canonical_to_branch`.
- `get_depth("LOC") == 2` (was `3`) — `LOC` now denotes the depth-2 branch.
- `CanonicalMapper.map()` no longer accepts them as resolution *targets*; the mapping is also no longer needed. Two `docs/mapping_scenarios.md` examples updated.
- `to_branch("LOC")` still returns `"LOCATION"`, unchanged.

## Tests

`TestBranchAliases` covers resolution, branch paths, reserved-key non-leakage at deep nodes, alias-as-subject, the `to_branch`/`get_depth` regression guards, non-default `canonical_depth` (2/3/4/10 — `CanonicalMapper` builds at depth 10), and the collision rollback.

**724 passed, 2 skipped**, ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@
 
 ## Version 0.3.2
 
+### Features
+
+- **Branch-level aliases** — non-leaf hierarchy nodes can now declare raw aliases via a reserved `_aliases` key (e.g. `"LOCATION": {"_aliases": ["LOC"], ...}`), mirroring the alias lists that leaf nodes already have. `add_alias()` on a branch node now records the alias instead of creating a spurious child leaf. The reserved key is skipped by every tree-walk, so it never becomes a canonical entity.
+
+### Breaking Changes
+
+- **`LOC` and `ORG` are no longer canonical entities** — they were empty leaf nodes under `LOCATION`/`ORGANIZATION` and are now branch-level aliases of them. Coarse dataset labels like TAB's `LOC`/`ORG` therefore match a model's `LOCATION`/`ORGANIZATION` at the exact (leaf) level, not only at the branch level. Concretely:
+  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise for `ORG`.
+  - `LOC`/`ORG` no longer appear in `all_canonical_entities` or `canonical_to_branch`.
+  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch.
+  - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
+  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged.
+
+### Behavior Changes
+
+- **`to_branch()` and `get_depth()` now resolve raw aliases**, not just canonical names. Previously a raw alias (e.g. `COMPANYNAME`, `QQ`) was passed through unchanged by `to_branch` and raised in `get_depth`; both now resolve it first. Unknown labels are still returned as-is by `to_branch`.
+- **`add_alias()` accepts an alias as its subject**, so `add_alias("LOC", ...)` works as well as `add_alias("LOCATION", ...)`.
+- **`add_alias()` now raises `ValueError` instead of silently no-opping** when the alias is already claimed by a descendant of the target (e.g. adding `CITY` to the `LOCATION` branch, where `CITY` already resolves to `ADDRESS`). The hierarchy is left unmodified. It also raises `KeyError` if the reserved `_aliases` key is passed as the entity name.
+
 ### Bug Fixes
 
 - **Span merging no longer depends on the DataFrame index** — `SpanEvaluator` mixed sentence-relative token positions with DataFrame index labels when checking whether two same-type spans are adjacent. With the global index produced by `predict_dataset()`, the between-tokens lookup read the wrong rows — or none at all — for every sentence except the one starting at row 0, and an empty lookup counts as "adjacent", silently merging same-type spans separated by regular words (e.g. the two PERSON spans in "John visited Berlin with Mary" became one). Span counts (`num_annotated`, `num_predicted`, `true_positives`) were deflated symmetrically for gold and predictions, so headline precision/recall could still look plausible. The evaluator now uses sentence-relative positions throughout and produces identical results for any DataFrame index.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,18 +10,19 @@
 
 ### Breaking Changes
 
-- **`LOC` and `ORG` are no longer canonical entities** — they were empty leaf nodes under `LOCATION`/`ORGANIZATION` and are now branch-level aliases of them. Coarse dataset labels like TAB's `LOC`/`ORG` therefore match a model's `LOCATION`/`ORGANIZATION` at the exact (leaf) level, not only at the branch level. Concretely:
-  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise for `ORG`.
-  - `LOC`/`ORG` no longer appear in `all_canonical_entities` or `canonical_to_branch`.
-  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch.
-  - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
-  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged.
+- **`LOC`, `ORG` and `PER` are no longer canonical entities** — they were empty leaf nodes under `LOCATION`/`ORGANIZATION`/`PERSON` > `NAME` and are now branch-level aliases of `LOCATION`/`ORGANIZATION`/`PERSON`. Coarse dataset labels like TAB's `LOC`/`ORG`/`PER` therefore match a model's `LOCATION`/`ORGANIZATION`/`PERSON` at the exact (leaf) level, not only at the branch level. Concretely:
+  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise for `ORG` and `PER`.
+  - `LOC`/`ORG`/`PER` no longer appear in `all_canonical_entities` or `canonical_to_branch`.
+  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch. `get_depth("PER")` returns `2` (was `3`).
+  - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG`/`PER` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
+  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged; `to_branch("PER")` still returns `"PERSON"`.
 
 ### Behavior Changes
 
 - **`to_branch()` and `get_depth()` now resolve raw aliases**, not just canonical names. Previously a raw alias (e.g. `COMPANYNAME`, `QQ`) was passed through unchanged by `to_branch` and raised in `get_depth`; both now resolve it first. Unknown labels are still returned as-is by `to_branch`.
 - **`add_alias()` accepts an alias as its subject**, so `add_alias("LOC", ...)` works as well as `add_alias("LOCATION", ...)`.
-- **`add_alias()` now raises `ValueError` instead of silently no-opping** when the alias is already claimed by a descendant of the target (e.g. adding `CITY` to the `LOCATION` branch, where `CITY` already resolves to `ADDRESS`). The hierarchy is left unmodified. It also raises `KeyError` if the reserved `_aliases` key is passed as the entity name.
+- **`add_alias()` now raises `ValueError` instead of silently no-opping** when the alias is already claimed by a descendant of the target (e.g. adding `CITY` to the `LOCATION` branch, where `CITY` already resolves to `ADDRESS`). The hierarchy is left unmodified — an alias the target already owns is preserved. It also raises `KeyError` if the reserved `_aliases` key is passed as the entity name.
+- **A branch alias shadowed by one of its own descendants logs a warning at construction time**, so collisions declared statically in `definitions.py` are no longer silent.
 
 ### Bug Fixes
 
--- a/docs/mapping_scenarios.md
+++ b/docs/mapping_scenarios.md
@@ -32,9 +32,9 @@ The model emits thousands of fine-grained labels; the dataset uses ≤10 broad c
 
 **Real example (Notebook 5):** The OpenMed HuggingFace model predicts `PERSON`, `LOCATION`, `ORGANIZATION`, etc., while the synth dataset labels `city`, `country`, `street_address`, `state`, `county`, `coordinate`, `postcode` as separate entities.
 
-**Issue type produced:** `COLLISION_AMBIGUOUS` (WARNING) — blocking. When a depth-2 ancestor like `LOCATION` is seen in predictions but the canonical surface is at depth 3 (computed by majority vote from the dataset annotations), the mapper can't determine which depth-3 entity to project onto. The user must call `map({'LOCATION': 'GEO'})` (or another appropriate target) to resolve.
+**Issue type produced:** `COLLISION_SAME_BRANCH` (INFO) — non-blocking. When a depth-2 ancestor like `LOCATION` is seen in predictions but the canonical surface is at depth 3 (computed by majority vote from the dataset annotations), the label and its co-occurring annotations sit on the same branch at different depths. Hierarchical evaluation handles this automatically by scoring at the branch level; no `map()` call is required. Use `map({'LOCATION': 'GEO'})` only if you want to force a specific depth-3 target.
 
-**Projection rules in action:** If instead the dataset uses depth-2 labels and the canonical surface locks at depth 2, then fine-grained model labels like `STREET_ADDRESS` auto-collapse to `LOCATION` as `COLLISION_TRIVIAL` (INFO, non-blocking).
+**Projection rules in action:** If instead the dataset uses depth-2 labels and the canonical surface locks at depth 2, then fine-grained model labels like `STREET_ADDRESS` auto-collapse to `LOCATION` — also reported as `COLLISION_SAME_BRANCH` (INFO, non-blocking).
 
 ---
 
@@ -112,7 +112,7 @@ The same string alias appears under multiple canonical entities in the hierarchy
 
 **Issue types produced:**
 - `COLLISION_CROSS_BRANCH` (WARNING) — blocking. Raised when a label resolves to a canonical entity that has co-occurring labels on the same tokens mapping to a different hierarchy branch. Must be resolved with `map()` before extracting results.
-- `COLLISION_AMBIGUOUS` (WARNING) — blocking. Raised when a depth-2 ancestor maps to multiple depth-3 entities on the canonical surface (the top co-occurring candidate is shown in `overlap_counts`). Use `map({'LABEL': 'CANONICAL'})` to pick the right one.
+- `COLLISION_SAME_BRANCH` (INFO) — non-blocking. Raised when a label co-occurs with annotations on the same branch at a different depth (e.g. prediction `PERSON` vs annotation `NAME`). Handled automatically by hierarchical evaluation; use `map({'LABEL': 'CANONICAL'})` only to force a specific target.
 
 ---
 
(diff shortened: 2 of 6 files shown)


presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -100,26 +107,75 @@ def get_depth(self, entity: str) -> int:
...
+        # than persist an alias that silently never resolves, roll back and say
+        # so — the caller's intent could not be honoured.
+        #
+        # The alias is correct when it resolves wherever the TARGET resolves:
+        # for a node below canonical_depth that is the canonical ancestor, not
+        # the node's own name.
+        expected = self.raw_to_canonical.get(self.normalize(key))
+        resolved = self.raw_to_canonical.get(self.normalize(alias))
+        if resolved != expected:
+            added_to.remove(alias)
```
negruber1: Rollback can delete a pre-existing alias. If `alias` is already in `added_to`, this call does not append it, but collision handling still executes `added_to.remove(alias)`. For example, `VRN` already belongs to both `LICENSE_PLATE_NUMBER` and `LICENSE_PLATE`; calling `add_alias("LICENSE_PLATE_NUMBER", "VRN")` raises and removes the former’s existing alias. I think that we should track whether this call appended the alias and only remove it during rollback when it did.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -100,26 +107,75 @@ def get_depth(self, entity: str) -> int:
...
+            if not added_to and added_to is not value:
+                value.pop(BRANCH_ALIASES_KEY, None)
+            self._rebuild()
+            raise ValueError(
+                f"Alias {alias!r} already resolves to {resolved!r}, "
+                f"so it cannot be added to {key!r}",
+            )
+            added_to.remove(alias)
```
negruber1: I think that lines 170-177 are dead code, because the raise exists the functions (it looks like an extra copy of the lines above)

presidio_evaluator/entity_mapping/definitions.py
```diff
@@ -87,6 +87,7 @@
             "SOCIAL_HANDLE": ["QQ"],  # QQ: Chinese messaging platform ID
         },
         "LOCATION": {
+            "_aliases": ["LOC"],
```
omri374, line 90: Please allso add "PER" to "PERSON"

docs/mapping_scenarios.md
```diff
@@ -32,7 +32,7 @@ The model emits thousands of fine-grained labels; the dataset uses ≤10 broad c
 
 **Real example (Notebook 5):** The OpenMed HuggingFace model predicts `PERSON`, `LOCATION`, `ORGANIZATION`, etc., while the synth dataset labels `city`, `country`, `street_address`, `state`, `county`, `coordinate`, `postcode` as separate entities.
 
-**Issue type produced:** `COLLISION_AMBIGUOUS` (WARNING) — blocking. When a depth-2 ancestor like `LOCATION` is seen in predictions but the canonical surface is at depth 3 (computed by majority vote from the dataset annotations), the mapper can't determine which depth-3 entity to project onto. The user must call `map({'LOCATION': 'LOC'})` (or another appropriate target) to resolve.
+**Issue type produced:** `COLLISION_AMBIGUOUS` (WARNING) — blocking. When a depth-2 ancestor like `LOCATION` is seen in predictions but the canonical surface is at depth 3 (computed by majority vote from the dataset annotations), the mapper can't determine which depth-3 entity to project onto. The user must call `map({'LOCATION': 'GEO'})` (or another appropriate target) to resolve.
```
omri374: Please also update entity_hierarchy.md with the new interfaces for alias and explanation on how they're used.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -162,6 +224,9 @@ def _build_alias_map(
                     mapping[EntityHierarchy.normalize(alias)] = key
             elif isinstance(value, dict):
                 mapping[EntityHierarchy.normalize(key)] = key
+                # Branch-level aliases: raw synonyms of this non-leaf node.
+                for alias in value.get(BRANCH_ALIASES_KEY, []):
```
omri374, line 224: A bit of an edge case: the collision guard only protects `add_alias(`). A collision baked statically into definitions.py - an _aliases entry whose normalized form matches a descendant is still silently shadowed, because the recursive `mapping.update()` overwrites the branch alias. Fine for LOC/ORG today; consider either a construction-time check in _rebuild() or a comment in `definitions.py` stating the invariant.

docs/mapping_scenarios.md
```diff
@@ -32,7 +32,7 @@ The model emits thousands of fine-grained labels; the dataset uses ≤10 broad c
...
 
 **Projection rules in action:** If instead the dataset uses depth-2 labels and the canonical surface locks at depth 2, then fine-grained model labels like `STREET_ADDRESS` auto-collapse to `LOCATION` as `COLLISION_TRIVIAL` (INFO, non-blocking).
```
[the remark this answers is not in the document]
RonShakutai: Fixed in c032311. Those two names no longer exist — swapped them for the real ones: `COLLISION_CROSS_BRANCH` (WARNING, blocks) and `COLLISION_SAME_BRANCH` (INFO, handled by hierarchical evaluation). Same fix applied in `entity_hierarchy.md`. Left ADR-003 alone since it records a decision at a point in time.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Good catch, fixed in c032311. `add_alias` now checks whether the alias was already present *before* appending, and only rolls back if this call actually added it. Added a test with the real case: `VRN` belongs to both `LICENSE_PLATE_NUMBER` and `LICENSE_PLATE`, and re-adding it used to raise *and* delete it.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Correct — that was a bad edit on my side, the block was duplicated. Removed in c032311.

presidio_evaluator/entity_mapping/definitions.py
RonShakutai: Done in c032311. Added `PER` as a branch alias of `PERSON`, same as `LOC`/`ORG`. It was an empty leaf under `PERSON > NAME`, so this is the same breaking change: `canonicalize("PER")` now returns `PERSON` and `get_depth("PER")` returns 2 instead of 3. CHANGELOG updated.

docs/mapping_scenarios.md
RonShakutai: Done in c032311. New **Aliases** section in `docs/entity_hierarchy.md` covering: leaf aliases vs branch aliases (`_aliases`), why coarse corpora like TAB need them, `add_alias()` (including passing an alias as the target), the collision `ValueError`, and the fact that `canonicalize`/`to_branch`/`get_depth` all accept raw aliases.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Agreed, fixed in c032311. Added a check that runs on every rebuild and logs a warning naming the alias, its branch, and what it actually resolves to — so a collision written straight into `definitions.py` is no longer silent. Chose a warning over an exception so existing custom hierarchies do not break on import. Verified the default hierarchy emits nothing, and added a test for both directions.

negruber1 (approved):

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -90,36 +100,78 @@ def get_branch(self, raw_label: str) -> list[str]:
...
 
+        # A branch alias is applied before the recursive descent into its own
+        # subtree, so a descendant with the same normalized name wins. Rather
+        # than persist an alias that silently never resolves, roll back and say
+        # so — the caller's intent could not be honoured.
+        #
+        # The alias is correct when it resolves wherever the TARGET resolves:
+        # for a node below canonical_depth that is the canonical ancestor, not
+        # the node's own name.
+        expected = self.raw_to_canonical.get(self.normalize(key))
```
omri374, line 162: `add_alias("LICENSE", "PROF_LIC")` worked on main and raises on this branch:

```py
from presidio_evaluator.entity_mapping.hierarchy import EntityHierarchy

h = EntityHierarchy()
h.add_alias("LICENSE", "PROF_LIC")

# main:    ok — canonicalize("PROF_LIC") == "LICENSE"
# this PR: ValueError: Alias 'PROF_LIC' already resolves to 'LICENSE',
#                      so it cannot be added to 'LICENSE'
```

`PROF_LIC` is a brand-new string. `h.normalize("PROF_LIC")` in `h.raw_to_canonical` is False before the call. It "already resolves to LICENSE" only because line 150 appended it and line 152 rebuilt the map, three lines above the check that reads it back. So the check isn't catching a pre-existing alias; it's catching the one it just added, landing exactly where the caller asked, and calling that a conflict.

presidio_evaluator/entity_mapping/definitions.py
```diff
@@ -11,6 +11,7 @@
 HIERARCHY: dict = {
     "PII": {
         "PERSON": {
+            "_aliases": ["PER"],
```
omri374: `PER` doesn't belong in this change — the PR's rationale doesn't apply to it, and moving it is a silent accuracy regression.

`LOC` and `ORG` really were stranded: empty leaves sitting directly on their branch, canonical entities in their own right, unable to ever match `LOCATION`/`ORGANIZATION` exactly. `PER` was not. It was a depth-4 leaf under `PERSON > NAME`, so it already canonicalized to `NAME` — the same place `FIRSTNAME`, `LASTNAME` and friends land:

| | `main` | this PR |
| --- | --- | --- |
| `canonicalize("PER")` | `'NAME'` | `'PERSON'` |
| `to_branch("PER")` | `'PER'` | `'PERSON'` |
| `get_depth("PER")` | raises `EntityNotMappedError` | `2` |

The practical effect: a corpus annotating `PER` evaluated against a model predicting `FIRSTNAME`/`LASTNAME` used to match at the **exact** level (both sides → `NAME`) and now mismatches (`PERSON` vs `NAME`). That's the opposite of what the PR sets out to do, and because `PERSON` is a legitimate resolution no error is raised — the scores just quietly drop.

Suggest dropping the `PER` hunks (this line and the `"PER": []` removal at old `definitions.py:40`) and keeping the PR to `LOC`/`ORG`. If `PER → PERSON` is genuinely wanted it deserves its own PR, since it's a semantic reclassification rather than a plumbing fix.

---

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@
...
+
+### Breaking Changes
+
+- **`LOC`, `ORG` and `PER` are no longer canonical entities** — they were empty leaf nodes under `LOCATION`/`ORGANIZATION`/`PERSON` > `NAME` and are now branch-level aliases of `LOCATION`/`ORGANIZATION`/`PERSON`. Coarse dataset labels like TAB's `LOC`/`ORG`/`PER` therefore match a model's `LOCATION`/`ORGANIZATION`/`PERSON` at the exact (leaf) level, not only at the branch level. Concretely:
+  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise for `ORG` and `PER`.
+  - `LOC`/`ORG`/`PER` no longer appear in `all_canonical_entities` or `canonical_to_branch`.
+  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch. `get_depth("PER")` returns `2` (was `3`).
+  - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG`/`PER` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
+  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged; `to_branch("PER")` still returns `"PERSON"`.
```
omri374, line 18: `LOC` and `ORG` were depth-3 canonical leaves, but `PER` was a depth-4 leaf under `PERSON > NAME`, so the "was" values don't carry over. Measured on `main` vs this branch:

| claim | actual `main` | actual PR |
| --- | --- | --- |
| "`canonicalize("PER")` returns `"PERSON"` (was `"PER"`)" | `'NAME'` | `'PERSON'` |
| "`PER` no longer appears in `all_canonical_entities`" | already absent | absent |
| "`get_depth("PER")` returns `2` (was `3`)" | raises `EntityNotMappedError` | `2` |
| "`to_branch("PER")` still returns `"PERSON"`" | `'PER'` — so it **did** change | `'PERSON'` |

The last row is the one worth flagging: it's listed under the reassuring "still returns … unchanged" phrasing, but `to_branch("PER")` is exactly the call whose behaviour changed. Anyone bucketing a `PER`-annotated corpus by branch gets a different answer after this upgrade and the CHANGELOG tells them not to look.

If the `PER` hunks come out of `definitions.py` (see the comment there), this whole section simplifies to `LOC`/`ORG`, where the stated values are all correct. If `PER` stays, the entry needs rewriting around "was `NAME`, now `PERSON`" and calling out the exact-match regression.

---

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -90,36 +100,78 @@ def get_branch(self, raw_label: str) -> list[str]:
...
+        resolved = self.raw_to_canonical.get(self.normalize(alias))
+        if resolved != expected:
+            if appended:
+                added_to.remove(alias)
+                if not added_to and added_to is not value:
+                    value.pop(BRANCH_ALIASES_KEY, None)
+                self._rebuild()
```
omri374, line 169: Nit: the rejection path emits a shadow warning for state that is then discarded. `_rebuild()` at line 152 runs `_warn_on_shadowed_branch_aliases()` while the doomed alias is still attached, so a failed `add_alias("LOCATION", "CITY")` logs

```
Branch alias 'CITY' on 'LOCATION' is shadowed by 'ADDRESS' and will never resolve to 'LOCATION'.
```

immediately before raising the `ValueError` that undoes it. One warning, not a loop. The rollback `_rebuild()` here is silent because the alias is already gone but it's noise about a hierarchy state that never persisted, and it reads as a complaint about `definitions.py` rather than about the caller's rejected argument.

Suppressing the warning during the probe rebuild (or only warning from `__init__`) keeps the log to statically-declared collisions, which is what the warning is for.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -317,6 +415,14 @@ def to_branch(self, label: str | None) -> str:
...
             return "O"
         branch_path = self.canonical_to_branch.get(label)
+        if branch_path is None:
+            # Not a canonical name — it may be a raw alias (leaf or branch
+            # level, e.g. "LOC"). Resolve it before giving up, so callers
+            # feeding raw dataset labels get the right branch instead of a
+            # silent pass-through into a different bucket.
+            canonical = self.raw_to_canonical.get(self.normalize(label))
+            if canonical is not None:
+                branch_path = self.canonical_to_branch.get(canonical)
```
omri374, line 425: This fix has a twin that didn't get it: `_to_l1()` in `presidio_evaluator/evaluation/base_evaluator.py:22` does the same depth-2 projection with the same direct lookup this method used to have, and no alias resolution:

```python
branch = hierarchy.canonical_to_branch.get(entity)
if branch is None:
    return entity  # unknown entity — pass through unchanged
```

On `main` the two are behaviourally identical — I ran both over all 570 labels in `raw_to_canonical | all_canonical_entities` and they agreed on every one. On this branch they disagree on **434 of 570**:

```python
h = EntityHierarchy()

_to_l1("LOC", h)      # 'LOC'       <- was 'LOCATION' on main
h.to_branch("LOC")    # 'LOCATION'

_to_l1("CITY", h)     # 'CITY'
h.to_branch("CITY")   # 'LOCATION'
```

Those two rows are different problems. `LOC`/`ORG` are an actual regression in `_to_l1`: they used to be canonical entities so the lookup hit, and now it misses and passes the raw label through, so a `LOC`-annotated span buckets into a `LOC` branch of its own instead of `LOCATION` — the exact mis-bucketing this PR exists to remove. The remaining ~430 are the pre-existing alias gap that this method just fixed and `_to_l1` didn't.

Nothing miscounts today, which is why this is a nit: `_to_l1` and `_to_l0` have no callers. `grep -rn '_to_l1'` turns up the definition and one mention in `tasks/prd-canonical-mapper-redesign-and-hierarchical-eval.md:371`, which describes them as the earlier names for what became `to_binary`/`to_branch`. So it's a landmine rather than a live bug — a diverged copy of the function this PR is fixing, one import away from any future branch-level evaluation.

My suggestion is deleting both helpers rather than teaching `_to_l1` about aliases, since `to_branch()`/`to_binary()` supersede them. Equally happy for that to be a separate cleanup PR if you'd rather keep this diff tight.

---

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: You're right, and sorry this went unanswered — I merged before reading this round. Fixed in #192.

Confirmed exactly as you described: `PROF_LIC` is absent from `raw_to_canonical` before the call, and the check was reading back the alias it appended three lines earlier.

Root cause is narrower than it looks: `LICENSE` is declared **twice** — a canonical leaf under `EMPLOYMENT` (definitions.py:166) and an alias of `PROFESSIONAL_LICENSE` (:241). So `canonicalize("LICENSE")` is `PROFESSIONAL_LICENSE`, and "where does the target's name resolve" stops being the same question as "where does the target live". The guard now derives the target from the node's **path** in the tree instead of a name lookup. Swept all 126 canonical entities: 1 rejected a fresh alias before, 0 now.

Chasing this also turned up a worse sibling: `add_alias("LOCATION", "EMAIL")` was *accepted* and silently re-pointed `EMAIL` from `EMAIL_ADDRESS` to `LOCATION`, on dict ordering alone. Also fixed.

CHANGELOG.md
RonShakutai: Correct on all four rows — I verified against `1663688` and reproduced your table exactly. Fixed in #192.

I'd wrongly assumed `PER` behaved like `LOC`/`ORG`; it was a depth-4 leaf under `PERSON > NAME`, so none of the "was" values carried over.

Kept the `PER` hunks and rewrote the section instead of dropping them, since `PER` is a label TAB actually uses. `to_branch("PER")` is now called out explicitly as a behaviour change — `PER` used to form a branch bucket of its own and now joins `PERSON` — rather than sitting under the "still returns … unchanged" phrasing.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Agreed, fixed in #192 — I took the option you suggested.

`_rebuild()` now takes `warn=False`, and `add_alias` passes it for both speculative rebuilds. Shadow warnings come only from construction, which is what they're for. Added `test_rejected_add_alias_does_not_warn` plus a control asserting construction *still* warns, so silencing the probe can't silently disable the real check.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Good catch on the landmine. Took your suggestion and deleted both, in #192, rather than teaching `_to_l1` about aliases.

Confirmed callerless — only reference outside the definition is the PRD line you found. `to_branch()`/`to_binary()` supersede them, and deleting means a future import can't resurrect the divergence.

Kept it in this PR rather than splitting it out, since it's four lines and directly the same defect class as the rest of the round.

PR #191, opened 2026-08-18
fix: stop span merging from collapsing unrelated entities at coarse levels

RonShakutai:
## Problem

`SpanEvaluator` decided both **where a span ends** (`_create_spans`) and **whether two spans merge** (`_merge_adjacent_spans`) by comparing the *visible* entity label. At the binary level `CanonicalMapper` rewrites every label to `PII`, so both checks were vacuous.

```
"Contact John Smith , 32 , john@example.com"
  detailed -> 3 spans (NAME / AGE / EMAIL_ADDRESS)
  binary   -> 1 span  ('John Smith 32 john@example.com')
```

The gold span count therefore depended on the granularity being scored. On ai4privacy (n=20): **25 at binary, 27 at branch, 29 at detailed** — against 29 in the source data. The three levels were each scored against a *different* ground truth, so they were not comparable to one another, which is the entire point of reporting them side by side.

Two things made this hard to notice:

- The distortion applied symmetrically to annotations and predictions, so precision/recall still looked plausible while ~14% of the ground truth had silently disappeared.
- It scores perversely: a model that correctly returns three separate entities is compared against one merged gold span and can fail the IoU test outright.

## Two independent mechanisms

Worth calling out, because fixing only the first leaves a residual (ai4privacy landed at 59/59/60):

1. `_merge_adjacent_spans` joins spans separated by skip words when their types match.
2. `_create_spans` groups **contiguous identical labels** into a single span. Two entities *touching* with no token between them (`"Ana Ruiz 29"`) were never two spans to begin with — merging never ran.

## Fix

`CanonicalMapper` now attaches `annotation_merge_key` / `prediction_merge_key` columns carrying the **finest-grained** label to every level of `MappedResults`. `SpanEvaluator` uses them to end a span and to decide a merge, so entities stay distinct however coarse the label being scored.

- Same-entity fragment merging (`"New"` + `"York"` → one `LOCATION`) is unchanged.
- DataFrames built **without** `CanonicalMapper` have no merge-key columns, fall back to `None`, and keep their previous behaviour. `test_without_merge_keys_behaviour_is_unchanged` pins this deliberately, so the fallback isn't mistaken for a complete fix.

## Verification

Gold counts now agree exactly at every level, on all three evaluation datasets:

| dataset | raw | binary | branch | detailed |
|---|---|---|---|---|
| ai4privacy n=20 | 65 | 60 | 60 | 60 |
| TAB n=20 | 1081 | 962 | 962 | 962 |
| n2c2 n=20 | 328 | 286 | 286 | 286 |

(`raw` > merged is expected and correct: genuine same-entity fragments.)

New `TestBinaryLevelOverMerge` (6 tests) covers both mechanisms. **734 passed, 2 skipped**; `ruff` clean.

## Known limitation (not addressed here)

Two *distinct* same-type entities separated by a skip word (`"She visited Paris , London"`) still merge into one, at **every** level — merging is meant for fragments of one entity, not neighbours that happen to share a type. Out of scope for this PR; filing separately.


presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -639,13 +731,14 @@ def _create_spans(self, df: pd.DataFrame, column: str) -> list[Span]:
...
                     current_entity_type = None
+                    current_merge_key = None
                     current_tokens = []
                     current_start_indices = []
                     current_token_start = 0
 
                 continue
 
-            if entity_type != current_entity_type:
+            if entity_type != current_entity_type or merge_key != current_merge_key:
```
omri374: Heads up on a side effect here: now that gold spans get split, a single prediction covering two gold spans gets counted once per gold span, because `_compare_single_overlaps` never checks `processed_predictions`. I ran the binary case with gold keys NAME NAME AGE on "John Smith 32" and one prediction over all three tokens: `pii_predicted=2`, `pii_false_positives=2` for a single predicted span. On main the same input scored TP=1, predicted=1. The counting bug is pre-existing, but this change makes it fire in exactly the adjacent-entity cases the PR is about, so precision denominators shift. Worth fixing here or in a fast follow.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -286,14 +311,70 @@ def _process_sentence_spans(
...
+        key_column = (
+            ANNOTATION_MERGE_KEY if column == "annotation" else PREDICTION_MERGE_KEY
+        )
+        if key_column not in sentence_df.columns:
+            return None
+
+        keys = []
+        for span in spans:
+            if span.token_start is None or span.token_start >= len(sentence_df):
+                return None
```
omri374: If one span has `token_start` None or out of range, this drops the merge keys for the whole sentence and merging falls back to comparing visible labels. At the binary level that's the vacuous PII == PII compare again, so the exact bug this PR fixes can quietly come back with no signal. `_merge_adjacent_spans` raises on a length mismatch, I'd raise (or at least warn) here too instead of degrading silently.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -286,14 +311,70 @@ def _process_sentence_spans(
...
+            PREDICTION_MERGE_KEY,
+        )
+
+        key_column = (
+            ANNOTATION_MERGE_KEY if column == "annotation" else PREDICTION_MERGE_KEY
+        )
+        return key_column if key_column in df.columns else None
+
+    @staticmethod
+    def _merge_keys_for(
```
omri374: Two things on this helper. It duplicates `_merge_key_column` almost verbatim (same import, same column resolution) instead of just calling it. And more broadly, `_create_spans` already knows each run's merge key and throws it away, then this re-derives it by indexing at `span.token_start`, which relies on that being an in-range positional index into the same df. If `_create_spans` returned the key alongside each span (or stashed it on the span), this helper and the None fallback above could go away entirely.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -87,23 +87,47 @@ def _normalize_tokens(
...
+        df: pd.DataFrame,
+        merge_keys: list[str] | None = None,
+    ) -> list[Span]:
         """
         Merge adjacent spans of the same entity type if separated only by skip words / punctuation.
 
         :param spans: List of Span objects to potentially merge
         :param df: DataFrame containing the tokens and their positions
+        :param merge_keys: Optional per-span finest-grained labels. When given,
+            these decide whether two spans describe the same entity type, instead
```
omri374: Tiny nit: the docstring says merge keys decide "instead of Span.entity_type", but the code below requires both the key and the entity type to match.

---

presidio_evaluator/entity_mapping/mapper.py
omri374: The `get_mapped_results_dataframe` docstring is stale now: it says each level carries annotation/prediction plus the original columns, and that `.original` is the raw input unmodified. All four DataFrames now also carry the two merge key columns, so anyone asserting the column set or round-tripping `.original` will trip on that. Worth a quick docstring update.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: This approach works and fixes the collapse, but I want to flag two limitations so we decide about them consciously.

First, a merge key is the finest-grained label of a token, so it can only distinguish entity types, not entity instances. That's enough for "John Smith, 32" at the binary level (NAME vs AGE), but two neighbouring entities of the same type share a key: in "She visited Paris , London" both carry LOCATION, so they still merge into one span at every level. That's the known limitation in the description, and it's inherent to using labels as the identity signal, no refinement of the keys can fix it.

Second, there's a structural cost. _create_spans reads each token's merge key while building spans but drops it, so this helper has to recover one key per span afterwards by indexing back into the df at span.token_start. That lookup depends on token_start being a valid positional index into the same sentence df, and when any span lacks it the whole sentence silently falls back to comparing visible labels, which at the binary level is the vacuous PII == PII compare again.

I think the cleaner place to determine span boundaries is span_to_tag. It's where spans get flattened into per-token labels, so it's the one place that already knows, for every token, which span produced its label. If it emits that span id alongside the tag, the evaluator receives entity instances instead of types: Paris is span 0 and London is span 1, so they never merge; "Ana Ruiz 29" with no separator splits exactly at the id change; and nothing needs re-deriving because the identity travels with the token from the source. The label-run inference then remains only as a fallback for inputs that genuinely lack span information, like token-tagging models that emit IO tags.

To make this concrete I asked Claude to put together a reference branch on top of yours, #202, so you can see what it looks like in practice. Take a look and tell me what you think. Happy to land it stacked on this PR or fold the ideas in here, whatever you prefer.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -645,13 +737,14 @@ def _create_spans(self, df: pd.DataFrame, column: str) -> list[Span]:
                             ),
                         )
                     current_entity_type = None
+                    current_merge_key = None
                     current_tokens = []
```
omri374: Heads up on a side effect here: now that gold spans get split, a single prediction covering two gold spans gets counted once per gold span, because _compare_single_overlaps never checks processed_predictions. I ran the binary case with gold keys NAME NAME AGE on "John Smith 32" and one prediction over all three tokens: pii_predicted=2, pii_false_positives=2 for a single predicted span. On main the same input scored TP=1, predicted=1. The counting bug is pre-existing, but this change (and #202 equally) makes it fire in exactly the adjacent-entity cases these PRs are about, so precision denominators shift. Worth fixing here or in a fast follow.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: If one span has token_start None or out of range, this drops the merge keys for the whole sentence and merging falls back to comparing visible labels. At the binary level that's the vacuous PII == PII compare again, so the exact bug this PR fixes can quietly come back with no signal. _merge_adjacent_spans raises on a length mismatch, I'd raise (or at least warn) here too instead of degrading silently. This survives in #202 as well, so wherever it gets handled works.

---

presidio_evaluator/entity_mapping/mapper.py
omri374: The get_mapped_results_dataframe docstring is stale now: it says each level carries annotation/prediction plus the original columns, and that .original is the raw input unmodified. All four DataFrames now also carry the two merge key columns (and the span id column once #202 lands), so anyone asserting the column set will trip on that. Worth a quick docstring update.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: This approach works and fixes the collapse, but I want to flag two limitations so we decide about them consciously.

First, the same-type case you list as a known limitation ("She visited Paris , London" merging at every level) isn't something a later refinement of the merge keys can fix. The keys are labels, and labels can tell a NAME from an AGE but can never tell two LOCATIONs apart, so instance identity is out of reach for this design no matter how fine-grained the keys get.

Second, there's a structural cost. _create_spans reads each token's merge key while building spans but drops it, so this helper has to recover one key per span afterwards by indexing back into the df at span.token_start. That lookup depends on token_start being a valid positional index into the same sentence df, and when any span lacks it the whole sentence silently falls back to comparing visible labels, which at the binary level is the vacuous PII == PII compare again.

I think the cleaner place to determine span boundaries is span_to_tag. It's where spans get flattened into per-token labels, so it's the one place that already knows, for every token, which span produced its label. If it emits that span id alongside the tag, the evaluator receives entity instances instead of types: Paris is span 0 and London is span 1, so they never merge; "Ana Ruiz 29" with no separator splits exactly at the id change; and nothing needs re-deriving because the identity travels with the token from the source. The label-run inference then remains only as a fallback for inputs that genuinely lack span information, like token-tagging models that emit IO tags.

To make this concrete I asked Claude to put together a reference branch on top of yours, #202, so you can see what it looks like in practice. Take a look and tell me what you think. Happy to land it stacked on this PR or fold the ideas in here, whatever you prefer.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: Heads up on a side effect here: now that gold spans get split, a single prediction covering two gold spans gets counted once per gold span, because _compare_single_overlaps never checks processed_predictions. I ran the binary case with gold keys NAME NAME AGE on "John Smith 32" and one prediction over all three tokens: pii_predicted=2, pii_false_positives=2 for a single predicted span. On main the same input scored TP=1, predicted=1. The counting bug is pre-existing, but this change (and #202 equally) makes it fire in exactly the adjacent-entity cases these PRs are about, so precision denominators shift. Worth fixing here or in a fast follow.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: If one span has token_start None or out of range, this drops the merge keys for the whole sentence and merging falls back to comparing visible labels. At the binary level that's the vacuous PII == PII compare again, so the exact bug this PR fixes can quietly come back with no signal. _merge_adjacent_spans raises on a length mismatch, I'd raise (or at least warn) here too instead of degrading silently. This survives in #202 as well, so wherever it gets handled works.

---

presidio_evaluator/entity_mapping/mapper.py
omri374: The get_mapped_results_dataframe docstring is stale now: it says each level carries annotation/prediction plus the original columns, and that .original is the raw input unmodified. All four DataFrames now also carry the two merge key columns (and the span id column once #202 lands), so anyone asserting the column set will trip on that. Worth a quick docstring update.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: This fixes the collapse, and the invariant test is the right check. Two limitations are worth stating because they bound what merge keys can ever do:

1. Keys are labels. They separate a NAME from an AGE but can never separate two LOCATIONs, so the "Paris , London" case in the description is not something a later refinement of the keys can reach. It needs instance identity, and the only place that has it is `span_to_tag`, where spans are flattened into per-token tags.

2. `_create_spans` reads each token's key while building spans and then drops it, so this helper rebuilds one key per span by indexing back into the sentence df at `token_start`. That works because `_create_spans` assigns positional indices, but it is a second source of truth for something the span could carry itself.

I put together #202 as a reference for the `span_to_tag` approach: it emits a per-token gold span id and carries it through `predict_dataset` into the evaluator. It is not ready to land. It fixes identity on the gold side only, while the evaluator still merges predictions across skip words, so a model that correctly returns "Paris" and "London" as two spans is scored 0 TP, 2 FP, 2 FN there, where this PR scores it 1 TP. Making it correct also means dropping prediction-side merging when gold ids exist (cumulative IoU already handles fragmented predictions), and touching same-type neighbours would still need BIO tags or spans from the model. None of that blocks this PR. Have a look and tell me what you think; this is a discussion, not a request for this PR.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -645,13 +737,14 @@ def _create_spans(self, df: pd.DataFrame, column: str) -> list[Span]:
...
                     current_start_indices = []
                     current_token_start = 0
 
                 continue
 
-            if entity_type != current_entity_type:
+            if entity_type != current_entity_type or merge_key != current_merge_key:
```
omri374: Side effect to be aware of: now that gold spans get split here, one prediction covering two gold spans is counted once per gold span, because `_compare_single_overlaps` never consults `processed_predictions`.

Binary level, gold NAME NAME AGE on "John Smith 32", one prediction over all three tokens:

- this branch: `pii_predicted=2`, `pii_false_positives=2` for a single predicted span
- main: `TP=1`, `predicted=1`, because the gold was merged into one span

The bug is pre-existing. Main already shows it at the detailed level for the same input (`NAME num_predicted=2`). But this change makes it fire in exactly the adjacent-entity cases the PR is about, so precision denominators shift on real datasets. Please fix it here, or in a follow-up that lands before these numbers are reported.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374: If any span has `token_start` None or out of range, the whole sentence silently drops its merge keys and merging falls back to comparing visible labels. At the binary level that is the vacuous `PII == PII` compare again, so the bug this PR fixes can quietly return with no signal.

`_create_spans` always sets a positional `token_start`, so this branch is unreachable today, which makes raising safe. `_merge_adjacent_spans` already raises on a length mismatch. Please do the same here instead of returning None.

---

presidio_evaluator/entity_mapping/mapper.py
omri374: The `get_mapped_results_dataframe` docstring is now stale. It says each level carries `annotation` and `prediction` plus the original columns, and that `.original` is the raw input unmodified. All four DataFrames now also carry `annotation_merge_key` and `prediction_merge_key`, so anyone asserting the column set will trip on that. Quick docstring update please.

---

presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -970,23 +972,38 @@ def _project_prediction(label: str) -> str:
...
+        # alongside every level so that span merging can tell entities apart even
+        # after their labels have been collapsed.
+        #
+        # Without this, adjacent-span merging at the binary level compares
+        # "PII" to "PII" and unconditionally merges neighbours, so a name, an age
+        # and an email separated by commas become one span. That silently removes
+        # gold spans and makes the levels incomparable, since each ends up scored
+        # against a different ground truth.
+        ann_merge_key = df["annotation"].map(lambda x: _level(x, "detailed"))
+        pred_merge_key = df["prediction"].map(_project_prediction)
```
omri374: Note on the two commits on your branch that you did not write.

My merge of main (954e720) resolved both conflict hunks in this function by keeping the branch side, which dropped the `_project_prediction` logic from #196 and left 15 tests failing. cf44d8c restores it: predictions are projected to the deepest annotated ancestor as before, the projected label becomes the prediction merge key, and each level collapses it with `to_binary` / `to_branch`, so the projection and the merge keys work together. ad92ee4 then merges main again to pick up the CI trigger fix (#203, #204) and Python 3.14 support (#205).

Please pull before pushing again. Also worth knowing: CI never ran pytest on PRs to main until today, so this push is the first real CI run on this PR. Locally, `pytest --runslow` on this head gives 755 passed.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374, line 346: Not a request for this PR, just a thought for later.

Merge keys are labels, so they tell a NAME from an AGE but not two LOCATIONs, which is the "Paris , London" case you already list as a known limitation. The one place that knows which span produced each token is `span_to_tag`, since that is where spans get flattened into per-token tags. If it also emitted the source span index per token, the evaluator could split on span identity instead of inferring boundaries from label runs, and this helper would not need to recover keys from the df afterwards.

I am not sure it is worth the plumbing. It only helps the gold side, and the prediction side has its own complications, so I would leave it as something to consider in a follow-up rather than change anything here. I sketched the shape in #202 if you are curious, but it is not ready and not part of this review.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374, line 747: Side effect to be aware of: now that gold spans get split here, one prediction covering two gold spans is counted once per gold span, because `_compare_single_overlaps` never consults `processed_predictions`.

Binary level, gold NAME NAME AGE on "John Smith 32", one prediction over all three tokens:

- this branch: `pii_predicted=2`, `pii_false_positives=2` for a single predicted span
- main: `TP=1`, `predicted=1`, because the gold was merged into one span

The bug is pre-existing. Main already shows it at the detailed level for the same input (`NAME num_predicted=2`). But this change makes it fire in exactly the adjacent-entity cases the PR is about, so precision denominators shift on real datasets. Please fix it here, or in a follow-up that lands before these numbers are reported.

---

presidio_evaluator/evaluation/span_evaluator.py
omri374, line 374: If any span has `token_start` None or out of range, the whole sentence silently drops its merge keys and merging falls back to comparing visible labels. At the binary level that is the vacuous `PII == PII` compare again, so the bug this PR fixes can quietly return with no signal.

`_create_spans` always sets a positional `token_start`, so this branch is unreachable today, which makes raising safe. `_merge_adjacent_spans` already raises on a length mismatch. Please do the same here instead of returning None.

---

presidio_evaluator/entity_mapping/mapper.py
omri374, line 1: The `get_mapped_results_dataframe` docstring is now stale. It says each level carries `annotation` and `prediction` plus the original columns, and that `.original` is the raw input unmodified. All four DataFrames now also carry `annotation_merge_key` and `prediction_merge_key`, so anyone asserting the column set will trip on that. Quick docstring update please.

---

presidio_evaluator/entity_mapping/mapper.py
omri374, line 985: Note on the two commits on your branch that you did not write.

My merge of main (954e720) resolved both conflict hunks in this function by keeping the branch side, which dropped the `_project_prediction` logic from #196 and left 15 tests failing. cf44d8c restores it: predictions are projected to the deepest annotated ancestor as before, the projected label becomes the prediction merge key, and each level collapses it with `to_binary` / `to_branch`, so the projection and the merge keys work together. ad92ee4 then merges main again to pick up the CI trigger fix (#203, #204) and Python 3.14 support (#205).

Please pull before pushing again. Also worth knowing: CI never ran pytest on PRs to main until today, so this push is the first real CI run on this PR. Locally, `pytest --runslow` on this head gives 755 passed.

---

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -286,14 +311,70 @@ def _process_sentence_spans(
...
+        """
+        from presidio_evaluator.entity_mapping.data_objects import (  # noqa: PLC0415
+            ANNOTATION_MERGE_KEY,
+            PREDICTION_MERGE_KEY,
+        )
+
+        key_column = (
+            ANNOTATION_MERGE_KEY if column == "annotation" else PREDICTION_MERGE_KEY
+        )
+        return key_column if key_column in df.columns else None
```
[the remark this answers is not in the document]
RonShakutai: Fixed in 3b5b6dc. Only the exact annotation and prediction columns use the associated merge keys; arbitrary columns return None. Regression tests cover pred_a, pred_b, and custom_gold with an unrelated prediction_merge_key present.

presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -87,23 +87,47 @@ def _normalize_tokens(
...
         """
         Merge adjacent spans of the same entity type if separated only by skip words / punctuation.
 
         :param spans: List of Span objects to potentially merge
         :param df: DataFrame containing the tokens and their positions
+        :param merge_keys: Optional per-span finest-grained labels. When given,
+            these decide whether two spans describe the same entity type, instead
+            of ``Span.entity_type``. This matters once labels have been collapsed:
+            at the binary level every span is ``"PII"``, so comparing
+            ``entity_type`` would merge a name, an age and an email into one span.
```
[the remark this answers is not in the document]
RonShakutai: Fixed in 3b5b6dc. The docstring now states that both the detailed scoring key and scored entity_type must match. Added a control proving identical keys cannot merge different scored labels.

presidio_evaluator/evaluation/span_evaluator.py
RonShakutai: Agreed: left source-span identity plumbing out of this PR, as requested. Documented the remaining Paris/London limitation explicitly rather than claiming label merge keys solve it. The proposal in #202 remains a separate follow-up.

presidio_evaluator/evaluation/span_evaluator.py
RonShakutai: Confirmed against the current published implementation, then fixed in 3b5b6dc. Your John Smith 32 example at IoU 1.0 produced predicted=2, FP=2; it now produces predicted=1, FP=1, FN=2. Each prediction is assigned to one annotation before scoring, preferring a qualifying same-type match then greatest IoU, with deterministic ties. This also avoids an earlier weak overlap consuming a later valid match and prevents two TPs at low thresholds. Existing per-gold same-type fragment aggregation remains. Regressions cover binary/branch/detailed, character/token IoU, reversed gold order, mixed overlap groups, confusion counts and error records. This is documented as local assignment, not global maximum matching. Full suite: 793 passed.

presidio_evaluator/evaluation/span_evaluator.py
RonShakutai: Fixed in 3b5b6dc. When paired metadata exists, None, negative, or out-of-range token_start raises ValueError instead of discarding the keys for the whole sentence. Parameterized tests cover both gold and prediction metadata, including a valid first span followed by an invalid span. Missing metadata still uses the intentional backward-compatible fallback.

presidio_evaluator/entity_mapping/mapper.py
RonShakutai: Updated in 3b5b6dc: get_mapped_results_dataframe and MappedResults now document both added metadata columns on all four frames. .original preserves input columns/values but has extra metadata; predictions use projected detailed scoring labels for the keys. End-to-end tests assert the column set, unchanged raw input, and restored ancestor projection.

presidio_evaluator/entity_mapping/mapper.py
RonShakutai: Pulled/fast-forwarded to the latest PR head 4747bf7 before applying and pushing 3b5b6dc. Preserved the restored _project_prediction logic and added an end-to-end check that FIRSTNAME/LASTNAME predictions against NAME gold produce NAME prediction merge keys. Full pytest --runslow now passes 793 tests. Thanks for catching the merge loss.

RonShakutai:
Integration follow-up: this head already contains current main (6db3769), so no redundant merge commit was needed. Verified the changelog retains upstream entries with one heading per Unreleased category, and the CI workflow exactly matches main. Reran the full slow suite: 793 passed. Current Python 3.11-3.14 pytest CI and CodeQL checks are green.

RonShakutai:
@omri374, could you please revisit this PR for another review? Your feedback has been addressed, with the implementation following your design direction. The same-type Paris/London merging limitation remains intentionally deferred, as discussed.

In addition to the regression suite and rubber-duck review, I ran a separate standalone acceptance script against commit 3b5b6dc (no pytest): all 38 expected-versus-actual checks passed. These cover all three evaluation levels, character/token IoU, exclusive counting of broad predictions, ancestor projection, missing/spurious predictions, sparse multi-sentence indices, and invalid metadata handling.

The review threads are left open for your verification. Thank you!

PR #192, opened 2026-08-18
fix: address review comments left unaddressed on #186

RonShakutai:
Follow-up to #186. @omri374 left a second review round on Aug 17; #186 was merged on Aug 18 without those comments being answered. My mistake — this addresses all four.

## 1. `add_alias()` rejected a valid alias on `LICENSE`

The rollback guard asked where the *target's name* resolves and compared that to where the new alias landed. Same question for every node but one: the hierarchy declares `LICENSE` both as a canonical leaf under `EMPLOYMENT` and as an alias of `PROFESSIONAL_LICENSE`, so `canonicalize("LICENSE") == "PROFESSIONAL_LICENSE"`.

```py
h.add_alias("LICENSE", "PROF_LIC")
# ValueError: Alias 'PROF_LIC' already resolves to 'LICENSE'
```

`PROF_LIC` was brand new — the mapping being reported as a conflict was the one the call created one line earlier, exactly as described in review. The guard now derives the target from the node's **path** in the tree, so another node claiming the same name can't confuse it.

Swept all 126 canonical entities: **1 rejected a fresh alias before, 0 now.**

## 2. `add_alias()` silently stole aliases from other entities

Found while fixing (1). `add_alias("LOCATION", "EMAIL")` was *accepted* and quietly re-pointed `EMAIL` from `EMAIL_ADDRESS` to `LOCATION` — invalidating any corpus annotated with that label, succeeding or failing purely on dict ordering. Pre-existing, not from #186, but the fix above already required reading the alias's prior owner, so it's a two-line check. Now refused before the hierarchy is touched.

## 3. Rejected `add_alias()` logged a spurious shadow warning

The speculative rebuild ran the shadow check while the doomed alias was still attached, so a failed call warned about a state discarded microseconds later, worded as though `definitions.py` were at fault. `_rebuild()` now takes `warn=False` for probe rebuilds; construction still warns.

## 4. `_to_l1()` / `_to_l0()` removed

Deleted rather than taught about aliases, per the review suggestion. `_to_l1` was a copy of the branch projection that never learned alias resolution, leaving it disagreeing with `to_branch()` on 434/570 labels — including `LOC`, bucketed under a `LOC` branch of its own instead of `LOCATION`, the exact mis-bucketing #186 exists to remove. Callerless, so nothing miscounted; deleting it means a future import can't resurrect the divergence.

## CHANGELOG correction

The #186 entry described `PER` as behaving like `LOC`/`ORG`. It didn't — `PER` was a depth-4 leaf under `PERSON > NAME`. Verified against `1663688` (the commit before #186); the review's table is correct on all four rows:

| claim in #186 | actual before | now |
|---|---|---|
| `canonicalize("PER")` was `"PER"` | `'NAME'` | `'PERSON'` |
| `PER` no longer in `all_canonical_entities` | already absent | absent |
| `get_depth("PER")` was `3` | raised `EntityNotMappedError` | `2` |
| `to_branch("PER")` "still returns `PERSON`" | `'PER'` — **it did change** | `'PERSON'` |

That last row is now called out as a behaviour change rather than filed under "unchanged".

## Verification

9 tests added. **7 fail on `2c61890`**; the other 2 are controls that must keep passing (construction still warns; `to_branch` resolves aliases). **737 passed, 2 skipped**; `ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- **`add_alias()` no longer rejects a valid alias on `LICENSE`** — the rollback guard added alongside branch aliases asked where the *target's name* resolves and compared that to where the new alias landed. Those are the same question for every node except one: the hierarchy declares `LICENSE` both as a canonical leaf under `EMPLOYMENT` and as an alias of `PROFESSIONAL_LICENSE`, so `canonicalize("LICENSE")` is `"PROFESSIONAL_LICENSE"` and a brand-new alias correctly attached to the `LICENSE` leaf looked like a conflict. `add_alias("LICENSE", "PROF_LIC")` raised `ValueError` claiming the alias "already resolves to 'LICENSE'" — describing the mapping the call had itself created one line earlier. The guard now derives the target from the node's path in the tree rather than from a name lookup, so it is unaffected by another node claiming the same name. Verified across all 126 canonical entities: previously 1 rejected a fresh alias, now none do.
+- **`add_alias()` no longer silently steals an alias from another entity** — `add_alias("LOCATION", "EMAIL")` was accepted and quietly re-pointed `EMAIL` from `EMAIL_ADDRESS` to `LOCATION`, invalidating every corpus already annotated with that label. Whether the theft succeeded depended only on dictionary ordering. An alias already resolving somewhere other than the requested target is now rejected before the hierarchy is touched.
+- **A rejected `add_alias()` no longer logs a spurious shadow warning** — the speculative rebuild ran the branch-alias shadow check while the doomed alias was still attached, so a failed call logged a warning about a hierarchy state that was discarded microseconds later, phrased as though `definitions.py` were at fault. Shadow warnings now come only from construction, which is what they are for.
+
+### Removals
+
+- **`_to_l1()` and `_to_l0()` removed from `base_evaluator`** — superseded by `EntityHierarchy.to_branch()` / `to_binary()`, and callerless. `_to_l1` was a copy of the branch projection that did not learn alias resolution when `to_branch()` did, leaving the two disagreeing on 434 of 570 labels — including `LOC`, which `_to_l1` bucketed under a `LOC` branch of its own rather than `LOCATION`, the exact mis-bucketing branch aliases exist to remove. Nothing called it, so nothing miscounted; it is deleted rather than fixed so the divergence cannot be reintroduced by a future import.
+
 ## Version 0.3.2
 
 ### Features
@@ -11,11 +21,14 @@
 ### Breaking Changes
 
 - **`LOC`, `ORG` and `PER` are no longer canonical entities** — they were empty leaf nodes under `LOCATION`/`ORGANIZATION`/`PERSON` > `NAME` and are now branch-level aliases of `LOCATION`/`ORGANIZATION`/`PERSON`. Coarse dataset labels like TAB's `LOC`/`ORG`/`PER` therefore match a model's `LOCATION`/`ORGANIZATION`/`PERSON` at the exact (leaf) level, not only at the branch level. Concretely:
-  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise for `ORG` and `PER`.
-  - `LOC`/`ORG`/`PER` no longer appear in `all_canonical_entities` or `canonical_to_branch`.
-  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch. `get_depth("PER")` returns `2` (was `3`).
+  `LOC` and `ORG` were depth-3 canonical leaves, but `PER` sat one level deeper, under `PERSON > NAME`, so it does not follow the same before/after as the other two. Taking them separately:
+  - `canonicalize("LOC")` returns `"LOCATION"` (was `"LOC"`), and likewise `canonicalize("ORG")`.
+  - `canonicalize("PER")` returns `"PERSON"` — **was `"NAME"`**, not `"PER"`.
+  - `LOC` and `ORG` no longer appear in `all_canonical_entities` or `canonical_to_branch`. `PER` never did, being below the canonical depth.
+  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch. `get_depth("PER")` returns `2` — it previously **raised `EntityNotMappedError`**.
   - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG`/`PER` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
-  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged; `to_branch("PER")` still returns `"PERSON"`.
+  - `to_branch("LOC")` returns `"LOCATION"`, unchanged.
+  - **`to_branch("PER")` returns `"PERSON"`, where it previously returned `"PER"`.** This one is a genuine behaviour change, not a no-op: anyone bucketing a `PER`-annotated corpus by branch gets a different answer after upgrading. `PER` used to fall through `to_branch` unresolved and form a bucket of its own; it now joins `PERSON`.
 
 ### Behavior Changes
 
--- a/presidio_evaluator/entity_mapping/hierarchy.py
+++ b/presidio_evaluator/entity_mapping/hierarchy.py
@@ -136,7 +136,7 @@ class EntityHierarchy:
                 found = self._find_node(canonical)
         if found is None:
             raise KeyError(f"Entity {entity_name!r} not found in hierarchy")
-        parent_dict, key = found
+        parent_dict, key, path = found
         value = parent_dict[key]
         if isinstance(value, list):
             added_to = value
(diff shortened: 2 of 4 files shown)


presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -136,39 +136,55 @@ def add_alias(self, entity_name: str, alias: str) -> None:
...
             raise KeyError(f"Entity {entity_name!r} not found in hierarchy")
-        parent_dict, key = found
+        parent_dict, key, path = found
         value = parent_dict[key]
         if isinstance(value, list):
             added_to = value
         else:
             # Branch (non-leaf) node: record the alias under the reserved key so
             # it maps to this branch, instead of creating a spurious child leaf.
             added_to = value.setdefault(BRANCH_ALIASES_KEY, [])
```
omri374, line 146: This `setdefault` runs before the new ownership guard, so a rejected `add_alias` on a branch node still mutates the tree:

```python
h = EntityHierarchy()
before = copy.deepcopy(h.hierarchy)

h.add_alias("EMPLOYMENT", "EMAIL")
# ValueError: Alias 'EMAIL' already resolves to 'EMAIL_ADDRESS',
#             so it cannot be added to 'EMPLOYMENT'

h.hierarchy == before                              # False
h.hierarchy["PII"]["EMPLOYMENT"]["_aliases"]       # []  <- left by the rejected call
```

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -298,7 +319,8 @@ def _rebuild(self) -> None:
             self.hierarchy,
             self.canonical_depth,
         )
-        self._warn_on_shadowed_branch_aliases()
+        if warn:
+            self._warn_on_shadowed_branch_aliases()
```
omri374, line 323: The method this now guards still contains the bug this PR fixes. `_warn_on_shadowed_branch_aliases` asks where the branch's *name* resolves:

```python
expected = self.raw_to_canonical.get(self.normalize(branch_key))
```

That's the same name-based lookup `add_alias` just replaced with `_canonical_name_for_path(path)`, and it goes wrong on the same shape, a branch whose name is also an alias of something else:

```python
hier = copy.deepcopy(HIERARCHY)
hier["PII"]["EMPLOYMENT"]["LICENSE"] = {"_aliases": ["WORK_LIC"], "SUB_LIC": []}

h = EntityHierarchy(hierarchy=hier)
# WARNING  Branch alias 'WORK_LIC' on 'LICENSE' is shadowed by 'LICENSE'
#          and will never resolve to 'PROFESSIONAL_LICENSE'.

h.canonicalize("WORK_LIC")   # 'LICENSE'  <- landed exactly where declared
```

The alias resolves correctly; the warning says it never will, and points at `PROFESSIONAL_LICENSE`, which has nothing to do with this branch. Note the warning is what tells maintainers a static collision exists, so a false positive here trains people to ignore the one signal for real collisions.

The shipped hierarchy emits 0 warnings today, so nothing is noisy right now. This only fires once someone adds a branch in that shape. `_collect_branch_aliases` would need to carry the path down for the fix, since `_canonical_name_for_path` needs it.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -136,39 +136,55 @@ def add_alias(self, entity_name: str, alias: str) -> None:
...
+                f"Alias {alias!r} already resolves to {previously!r}, "
+                f"so it cannot be added to {key!r}",
+            )
         # Track whether THIS call appended, so a rollback never deletes an alias
         # that was already there (e.g. re-adding an alias the target already owns).
         appended = alias not in added_to
         if appended:
             added_to.append(alias)
-        self._rebuild()
+        self._rebuild(warn=False)
```
omri374, line 166: The `warn=False` change isn't covered — reverting both call sites to `self._rebuild()` leaves the suite green:

```
$ sed -i 's/self\._rebuild(warn=False)/self._rebuild()/g' hierarchy.py
$ pytest tests/entity_mapping -q
501 passed
```

`test_rejected_add_alias_does_not_warn` uses `add_alias("LOCATION", "CITY")`, which the new ownership guard rejects at line 152 — before line 166 ever runs. No rebuild, so no warning, with or without `warn=False`. It passes for a reason unrelated to what it's named after.

This is really the same observation as the dead-guard comment: nothing can currently reach the probe rebuild *and* produce a shadow warning, which is why the flag is untestable through the public API. If the second guard goes, `warn=False` can go with it. If it stays as a backstop, the honest test is a direct `_rebuild(warn=False)` call asserting no warning, rather than one routed through `add_alias`.

Its sibling `test_construction_still_warns_about_static_shadowing` is a genuine control and does fail if `_rebuild`'s default flips to `False` — worth keeping either way.

---

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -136,39 +136,55 @@ def add_alias(self, entity_name: str, alias: str) -> None:
...
-        # The alias is correct when it resolves wherever the TARGET resolves:
-        # for a node below canonical_depth that is the canonical ancestor, not
-        # the node's own name.
-        expected = self.raw_to_canonical.get(self.normalize(key))
+        # "Where the target lives" is computed from the node's PATH, not by
+        # asking where its name resolves. Those differ whenever another node
+        # claims the same name as an alias: the hierarchy has a canonical leaf
+        # LICENSE and also lists LICENSE as an alias of PROFESSIONAL_LICENSE, so
+        # raw_to_canonical["LICENSE"] is "PROFESSIONAL_LICENSE" and a name-based
+        # check rejected brand-new aliases that had landed exactly where asked.
```
omri374, line 178: LICENSE is a bug unrelated to this (I opened #199), so let's not use it as an example.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -136,39 +136,55 @@ def add_alias(self, entity_name: str, alias: str) -> None:
...
         resolved = self.raw_to_canonical.get(self.normalize(alias))
         if resolved != expected:
```
omri374, line 180: This can't fire any more. The check at line 152 already covers it.

Line 152 refuses when the alias name is already owned by someone other than the target, so there are only two ways past it: nobody owns the name (`previously is None`), or the target already owns it
(`previously == expected`).

If the target already owns it, adding it again changes nothing, so `resolved == expected`. If nobody owns it, this call adds the only occurrence, on the target, so again `resolved == expected`. Either way this
condition is false.

Being shadowed by a descendant means that descendant already owns the name,
which is the case line 152 rejects. Both checks ask the same question, one
before the write and one after.

The example in the comment above shows it. `add_alias("LOCATION", "CITY")` stops at line 152 with `already resolves to 'ADDRESS'` and never gets here.

Every node name in the tree is in `raw_to_canonical`, deep leaves included, so no descendant can shadow you without line 152 seeing it first.

So this block, its message, the rollback and the second `_rebuild` are dead, and `warn=False` only exists to keep that rollback quiet.

Delete it, or keep it as a backstop with a note that nothing reaches it. As written, the comment above reads like this is the main guard.

presidio_evaluator/entity_mapping/hierarchy.py
```diff
@@ -136,39 +136,55 @@ def add_alias(self, entity_name: str, alias: str) -> None:
...
+            # LOCATION, depending only on dict ordering.
+            raise ValueError(
+                f"Alias {alias!r} already resolves to {previously!r}, "
+                f"so it cannot be added to {key!r}",
+            )
         # Track whether THIS call appended, so a rollback never deletes an alias
         # that was already there (e.g. re-adding an alias the target already owns).
         appended = alias not in added_to
         if appended:
             added_to.append(alias)
```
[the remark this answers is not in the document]
RonShakutai: Adopted in 09f2d4d. Alias-list idempotency now compares normalized forms, after ownership validation. Tests cover leaf and branch targets with case/underscore/dash variants, and manual checks cover 451 fresh/repeated aliases across five canonical depths.

CHANGELOG.md
```diff
@@ -11,11 +21,14 @@
...
+  - `canonicalize("PER")` returns `"PERSON"` — **was `"NAME"`**, not `"PER"`.
+  - `LOC` and `ORG` no longer appear in `all_canonical_entities` or `canonical_to_branch`. `PER` never did, being below the canonical depth.
+  - `get_depth("LOC")` returns `2` (was `3`), because `LOC` now denotes the depth-2 `LOCATION` branch. `get_depth("PER")` returns `2` — it previously **raised `EntityNotMappedError`**.
   - `CanonicalMapper.map()` no longer accepts `LOC`/`ORG`/`PER` as resolution *targets*, since targets must be canonical entities. Such mappings are also no longer needed — the labels resolve on their own.
-  - `to_branch("LOC")` still returns `"LOCATION"`, unchanged; `to_branch("PER")` still returns `"PERSON"`.
+  - `to_branch("LOC")` returns `"LOCATION"`, unchanged.
+  - **`to_branch("PER")` returns `"PERSON"`, where it previously returned `"PER"`.** This one is a genuine behaviour change, not a no-op: anyone bucketing a `PER`-annotated corpus by branch gets a different answer after upgrading. `PER` used to fall through `to_branch` unresolved and form a bucket of its own; it now joins `PERSON`.
```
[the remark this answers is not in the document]
RonShakutai: Corrected to behavior in 09f2d4d.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Reproduced exactly: rejected EMPLOYMENT/EMAIL left an empty _aliases list. Fixed in 09f2d4d by checking ownership before setdefault. Regression tests compare the complete hierarchy object state and derived maps, not just raw_to_canonical, and verify rejection performs no rebuild.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Confirmed the false warning from your example and fixed in 09f2d4d. The collector carries branch paths and the warning derives its target with _canonical_name_for_path. Also found the inverse: if the alias and branch name both resolve to the other node, the old check hides a real collision. Permanent tests use a small synthetic hierarchy and cover both directions at multiple depths. Your exact example was also manually rerun: correct resolution and no warning.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Agreed. Instrumentation confirmed the rejected call performs zero rebuilds, so the old test could not establish warning-flag coverage. In 09f2d4d I took your simplification route: deleted the redundant guard and warn flag, moved static checks to construction, and renamed/strengthened the rejection test to assert zero rebuild calls. Kept construction controls and added genuine and false static-collision tests plus a successful-addition control.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Adopted in 09f2d4d. Removed the LICENSE example from implementation comments, the changelog rationale, this PR description, and permanent regression fixtures. Tests now use a synthetic custom hierarchy with a branch name reused as another node's alias. No taxonomy changes from #199 are included.

presidio_evaluator/entity_mapping/hierarchy.py
RonShakutai: Validated the argument against _build_alias_map and the raw-label coverage at canonical depths 1/2/3/4/10: node names and deep descendants are present before addition, so the pre-write ownership check sees those conflicts. In 09f2d4d I deleted the post-write guard, rollback, and second rebuild instead of retaining a misleading backstop. Removed the warn parameter too; static warnings run at construction. Fresh aliases rebuild once; rejected additions do not rebuild. Full suite passes 753 tests (including an author-approved repair of the pre-existing PATIENT test setup).

RonShakutai:
Merged current main (6db3769) in 1835232. Resolved both conflicts without discarding either side: preserved all upstream and PR-specific test methods (including LICENSE/PATIENT regression coverage), consolidated changelog sections, and adopted main-targeted CI with Python 3.11-3.14. Full local slow suite: 771 passed; targeted Ruff checks passed. The fresh GitHub pytest matrix is now being verified.

RonShakutai:
CI verification complete for integration head 1835232: pytest jobs on Python 3.11, 3.12, 3.13, and 3.14 all passed, as did CodeQL. GitHub reports MERGEABLE. Local full slow suite: 771 passed. Main integration, changelog consolidation, and preservation of both regression sets are complete; ready for reviewer re-review.

RonShakutai:
@omri374, could you please revisit this PR for another review? Your feedback has been addressed: alias ownership is validated before mutation, static warnings use structural paths, redundant rollback/warning controls were removed, and the tests now exercise the relevant behavior. The PR has also been integrated with main.

In addition to the regression suite, I ran a separate standalone acceptance script in this PR’s isolated checkout at commit 1835232 (no pytest): all 19 checks passed, including a sweep of 448 alias cases across five hierarchy depths. These verify atomic rejection, valid customization, normalized idempotency, real versus false collision warnings, instance isolation, and preservation of the upstream LICENSE/PATIENT/PER behavior.

The review threads are left open for your verification. Thank you!

PR #196, opened 2026-09-01
Credit more-specific hierarchy predictions

RonShakutai:
## Summary

Hierarchical detailed scoring currently treats a more-specific prediction as mislabeled, for example `gold=PERSON, pred=NAME` and `gold=DATE_TIME, pred=DATE`.

This revision implements descendant credit entirely in `CanonicalMapper`, keeping span and token evaluators hierarchy-agnostic.

**The rule: the gold annotation vocabulary decides the granularity, per prediction.** Each prediction is projected to the deepest annotation label that is an ancestor-or-self of *its own label*; a prediction with no annotated ancestor is left unchanged.

Because the walk follows the prediction's own ancestor chain, the consequences fall out for free:

- A more-specific prediction is credited to the gold label it falls under (`NAME` → `PERSON`, `DATE` → `DATE_TIME`).
- A coarser prediction is never pushed downward and remains a detailed-level mismatch.
- Siblings are never conflated — `DATE` cannot credit a `TIME` gold.
- **Mixed annotation depths on one branch need no decision.** With `PERSON` (depth 2) and `TITLE` (depth 3) both annotated, a `TITLE` prediction stays `TITLE` and a `NAME` prediction becomes `PERSON`, so every annotated depth keeps its own metrics. This is reported as an INFO issue only.
- Low-IoU descendant errors are attributed to the projected scoring label.
- Custom hierarchies and the legacy direct `TokenEvaluator` generic-entity behavior remain supported.

`data/synth_dataset_v2.json` — the dataset behind notebooks 4 and 5, which annotates PERSON (857 spans) and TITLE (92 spans) — maps with zero blocking issues and no `map()` call, with TITLE preserved in `detailed.per_type`.

Also in this PR:

- `SpanEvaluator.calculate_score_on_df` declares `allow_generic_entities` explicitly instead of swallowing it via `**kwargs`, documenting that span matching compares entity types exactly and has no generic-entity shortcut.
- Fixes a latent bug where the full-depth projection view was built from a module-level default hierarchy, so a mapper constructed with a custom `EntityHierarchy` projected against the built-in taxonomy.
- Notebooks 4, 5 and 6, the mapping documentation, ADR-002, ADR-003, the changelog, the renderer, and the notebook integration test were updated for the new rule.

## Validation

- Full suite: 745 passed, 2 skipped
- Ruff format and check clean
- 9-row behavior table verified against the live hierarchy (`PERSON ← NAME`, `DATE_TIME ← DATE`, coarser mismatches, sibling separation, and every PERSON/TITLE combination)
- Manual end-to-end scenario over the real `synth_dataset_v2` annotations: 0 blocking issues, TITLE present in both detailed gold and predictions, detailed recall PERSON=1.0 and TITLE=1.0
- Rubber-duck review completed; findings addressed

## Current behavior

`DATE_TIME` annotations are reported as mislabeled when the model predicts the more-specific `DATE`:

DATE_TIME annotation with DATE prediction

`PERSON` annotations are reported as mislabeled when the model predicts the more-specific `NAME`:

PERSON annotation with NAME prediction


presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -607,7 +628,126 @@ def _branch_key(resolved: str | None) -> str | None:
...
+                    type=IssueType.COLLISION_SAME_BRANCH,
+                    severity=IssueSeverity.ERROR,
+                    message=(
+                        f"Annotations on the {branch!r} branch mix hierarchy depths: "
+                        f"{resolved_depths}. Detailed predictions have no single "
+                        "projection target. Align these labels to one canonical depth "
+                        "before evaluation."
+                    ),
+                    labels=labels,
+                    annotation_count=sum(
```
omri374: **This new blocking error removes evaluation capability that previously worked correctly.**

To be clear about what this is *not*: given the current "one depth per branch" projection rule, blocking here is honest rather than buggy. I checked whether any single global rename could handle a mixed-depth branch and none can — `target_depth=2` breaks correct `TITLE` predictions, `target_depth=3` breaks `NAME`-on-`PERSON` credit. So the error correctly refuses a case the rule can't express. My concern is the scope of what it refuses.

**1. It blocks ADR-003's own motivating example.** The ADR's central case is *"a dataset that annotates `TITLE` and `PERSON` (PERSON branch)"*, with a table of expected per-level outcomes. That dataset now raises `IncompleteMapping` and cannot be evaluated at all.

**2. Neither resolution option preserves the measurement.** Running the ADR's Model A (all-`PERSON`, which the ADR says must be penalized at detailed) against Model B (perfect):

| resolution | outcome |
|---|---|
| `map({'TITLE': 'PERSON'})` | Both models produce **identical** detailed frames — the level can no longer distinguish a model that finds titles from one that doesn't |
| `map({'PERSON': None})` | The `PERSON` annotations are dropped entirely; `NAME` predictions on them become FPs |

**3. It is not TITLE-specific.** The same depth-2 + depth-3 collision exists on 18 branches — `LOCATION` + `ADDRESS`, `CONTACT` + `EMAIL_ADDRESS`, `ORGANIZATION` + `COMPANY`, `DATE_TIME` + `BIRTH_DATE` (plausible in medical data: most dates as `DATE_TIME`, birth dates specifically). Under `PERSON` alone it's `ALIAS`, `NAME`, `PREFIX`, `SUFFIX`, `TITLE`, `USERNAME`. So this isn't a taxonomy wart to fix by moving one node.

**4. Scenario 3 in the test suite had no bug.** It previously had three passing tests asserting depth-2 golds TP and depth-3 golds FN against depth-2 predictions — exactly what ADR-003 prescribes. Those are now replaced by a single test asserting the block.

The `_project_prediction` comment suggests a rule that removes the need for this error entirely. If you'd rather keep the block, then ADR-003's example table should be updated, since it currently documents expected outputs the system can no longer produce, and the CHANGELOG should call out that datasets with mixed gold depths hit a hard stop on upgrade.

---

presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -121,6 +117,30 @@ def pending(self) -> list[str]:
             lbl for lbl, rec in self._records.items() if rec.tier == "UNRESOLVED"
         )
 
+    @property
+    def _full_hierarchy(self) -> EntityHierarchy:
+        """Return a full-depth view of the mapper's current hierarchy."""
+        return EntityHierarchy(
+            hierarchy=self._hierarchy.hierarchy,
+            canonical_depth=10,
+        )
```
omri374: Nice catch — this fixes a latent bug beyond the stated scope of the PR. The old module-level `_FULL_HIERARCHY` was built from the *default* hierarchy, so a `CanonicalMapper` constructed with a custom hierarchy silently got wrong branch lookups in issue detection and level projection. I verified a custom taxonomy now resolves correctly end to end. Worth a line in the CHANGELOG under Bug Fixes.

Minor: this is a property rather than a constant now, so it rebuilds an `EntityHierarchy` on each access (~0.7 ms). Both call sites hoist it into a local so the cost is negligible — just flagging it so nobody later calls it inside a loop. A `cached_property` would need invalidation on `add_alias()` / `map()`, so keeping it live is probably the right call; a short comment saying so would help.

---

presidio_evaluator/evaluation/base_evaluator.py
```diff
@@ -300,7 +282,19 @@ def calculate_hierarchical_scores(
             )
 
         return {
-            "binary": self.calculate_score_on_df(results.binary, beta=beta),
-            "branch": self.calculate_score_on_df(results.branch, beta=beta),
-            "detailed": self.calculate_score_on_df(results.detailed, beta=beta),
+            "binary": self.calculate_score_on_df(
+                results.binary,
+                beta=beta,
+                allow_generic_entities=False,
```
omri374: Minor: `SpanEvaluator.calculate_score_on_df()` has no `allow_generic_entities` parameter, so this is silently swallowed by its `**kwargs` on all three calls. It works today because the span evaluator never had generic-entity forgiveness, but a caller reading this would reasonably assume the flag does something for spans, and a future `**kwargs` tightening would turn it into a `TypeError`.

Either accept and ignore it explicitly in `SpanEvaluator` with a short docstring note, or add a comment here saying it applies to token scoring only.

Gating this is the right call regardless of where projection lives, by the way: `gold=PERSON, pred=PII` can't be projected downward, so those rows reach the evaluator mismatched and the legacy forgiveness would otherwise flip them into TPs at the detailed level.

---

docs/adr/ADR-003-hierarchical-evaluation.md
```diff
@@ -2,12 +2,28 @@
...
+`SpanEvaluator` and `TokenEvaluator` independent of the hierarchy:
+
+- More-specific predictions are projected upward to the single annotation depth used
+  on their hierarchy branch.
+- Less-specific predictions are not projected downward and remain detailed-level
+  mismatches.
+- Mixed annotation depths on one branch are a blocking mapping issue requiring an
+  explicit `map()` decision.
+- Low-IoU errors are attributed to the projected scoring label.
+
```
omri374: Marking the ADR Superseded and recording the new design here is good practice — thanks for doing it rather than silently diverging.

One gap: the retained "original proposal" below still contains the worked example that motivates the whole ADR — a dataset annotating `TITLE` and `PERSON` on the PERSON branch — together with a table of expected per-level outcomes for Model A and Model B. Under the superseding implementation that dataset raises `IncompleteMapping` and cannot be evaluated, so the document now documents outputs the system can no longer produce.

If the blocking behavior stays, please either update that example to a single-depth vocabulary, or add a line to this section noting explicitly that the example below is no longer supported and why. If the projection rule changes per my comment on `_project_prediction`, the example works again as written and only this section needs adjusting.

---

notebooks/4_Evaluate_Presidio_Analyzer.ipynb
```diff
@@ -462,6 +462,18 @@
...
+    "for issue in mapper.get_issues():\n",
+    "    if issue.type.value == \"collision_same_branch\" and issue.severity.value == \"error\":\n",
+    "        resolutions.update(issue.resolution_options[0].mapper_call)\n",
+    "    elif issue.type.value in (\"unresolved\", \"prediction_only\"):\n",
+    "        resolutions.update(dict.fromkeys(issue.labels))\n",
+    "if resolutions:\n",
+    "    print(\"Applying explicit mappings before evaluation:\", resolutions)\n",
+    "    print(\"Review collapsed labels: detailed categories may be merged.\")\n",
+    "    mapper.map(resolutions)\n",
+    "\n",
```
omri374: This cell is required for the notebook to run at all after this change — `synth_dataset_v2.json` annotates both `PERSON` and `TITLE`, so `get_mapped_results_dataframe()` raises `IncompleteMapping` without it.

Auto-applying `resolution_options[0]` picks `{'TITLE': 'PERSON'}`, which collapses all 92 TITLE annotations into PERSON before scoring. `TITLE` then disappears from `detailed.per_type` entirely, so the notebook that exists to demonstrate hierarchical evaluation demonstrates it with the PERSON-branch granularity distinction switched off — a model that finds honorifics and one that labels them all `PERSON` score identically.

The printed warning is good, but blindly taking `resolution_options[0]` is also a pattern worth not teaching: it's the option that silently merges categories, and a reader copying this into their own pipeline inherits that. If the projection rule changes as suggested on `_project_prediction`, this whole cell can be removed and the notebook goes back to evaluating TITLE on its own terms.

---

docs/adr/ADR-003-hierarchical-evaluation.md
omri374: Marking the ADR Superseded and recording the new design here is good practice — thanks for doing it rather than silently diverging.

One thing to reconcile: the retained "original proposal" below contains the worked example that motivates the whole ADR — a dataset annotating `TITLE` and `PERSON` on the PERSON branch — with a table of expected per-level outcomes for Model A and Model B. That is not a hypothetical; it is `data/synth_dataset_v2.json`, the dataset NB4 and NB5 run on. Under the superseding implementation it raises `IncompleteMapping`, so this document now specifies behavior the system cannot produce on the repo's own primary dataset.

I don't think the fix is to rewrite the example — it's a reasonable annotation scheme and the ADR is right about how it should score. The `_project_prediction` comment proposes a rule under which this example works exactly as the table below states. If that lands, this section just needs its third bullet ("Mixed annotation depths on one branch are a blocking mapping issue") updated.

---

tests/entity_mapping/test_hierarchical_evaluation.py
```diff
@@ -273,73 +376,30 @@ def test_scenario2_branch_person_annotations_miss_pii_predictions(self):
...
-            0.0, abs=1e-6
-        )
-        assert scores["detailed"].per_type["GPE"].recall == pytest.approx(0.0, abs=1e-6)
+    def test_scenario3_mixed_gold_depths_block_mapping(self):
+        """S3 cannot choose one detailed projection depth for either branch."""
+        with pytest.raises(IncompleteMapping, match="blocking issue"):
+            _make_results(
+                ["PERSON", "LOCATION", "NAME", "GPE"],
+                ["PERSON", "LOCATION", "PERSON", "LOCATION"],
+            )
```
omri374: This replaces three tests that asserted *correct* behavior, not buggy behavior. On `main`, Scenario 3 (`PERSON, LOCATION, NAME, GPE` vs `PERSON, LOCATION, PERSON, LOCATION`) had `test_scenario3_binary_is_perfect`, `test_scenario3_branch_all_tp_because_depth3_maps_to_same_branch`, and `test_scenario3_detailed_depth2_annotations_hit_depth3_annotations_miss` — the last asserting depth-2 golds are TPs and depth-3 golds are FNs against depth-2 predictions, which is exactly what ADR-003 prescribes for a coarser model.

So this scenario was unaffected by the bug the PR fixes; it went from correctly scored to un-evaluable, and the coverage it provided (that a coarser model is penalized at detailed but not at branch) is now gone rather than relocated. Under the rule suggested on `_project_prediction` all three original assertions hold unchanged, so this would be a straight revert of this test back to the three.

Also worth adding whichever way this goes: a test pinning the gold vocabulary of `data/synth_dataset_v2.json` (or at least a `PERSON` + `TITLE` case) through `get_mapped_results_dataframe()`, since that combination is what NB4/NB5 actually exercise and nothing in the unit suite currently covers it end to end.

---

presidio_evaluator/evaluation/base_evaluator.py
omri374: Minor: `SpanEvaluator.calculate_score_on_df()` has no `allow_generic_entities` parameter, so this is silently swallowed by its `**kwargs` on all three calls. It works today because the span evaluator never had generic-entity forgiveness, but a reader would reasonably assume the flag does something for spans, and any future tightening of `**kwargs` turns it into a `TypeError`.

Either accept and ignore it explicitly in `SpanEvaluator` with a docstring note, or add a short comment here saying it only affects token scoring.

Gating this is the right call regardless of where projection ends up living, by the way: `gold=PERSON, pred=PII` can't be projected downward, so those rows reach the evaluator mismatched and the legacy forgiveness would otherwise flip them into TPs at the detailed level.

---

presidio_evaluator/entity_mapping/mapper.py
omri374: I'll let Claude explain as he does it better than I do :)


This blocks `data/synth_dataset_v2.json`, the dataset both NB4 and NB5 use.

Its gold vocabulary contains `PERSON` (857 spans) and `TITLE` (92 spans), which is exactly ADR-003's motivating example. Running this branch's mapper on that vocabulary:

```
[COLLISION_SAME_BRANCH] Annotations on the 'PERSON' branch mix hierarchy depths:
  PERSON (depth 2), TITLE (depth 3).
  -> map_to_canonical: {'TITLE': 'PERSON'}
  -> suppress: {'PERSON': None}
*** IncompleteMapping raised ***
```

So the `resolution_options[0]` boilerplate added to both notebooks and both integration tests in this PR is not defensive, it is load-bearing. And `resolution_options[0]` is `{'TITLE': 'PERSON'}`, which means the repo's flagship demo now silently folds all 92 TITLE annotations into PERSON. I confirmed `TITLE` disappears from `detailed.per_type` entirely afterwards: a model that correctly identifies honorifics and one that labels every honorific `PERSON` produce identical detailed scores. The notebook whose purpose is to demonstrate hierarchical evaluation would ship with the granularity distinction switched off.

To be fair about what this is not: given the current "one depth per branch" rule, blocking here is honest rather than buggy. I checked whether any single global rename can handle a mixed-depth branch and none can. Using `target_depth=2` breaks correct `TITLE` predictions, and `target_depth=3` breaks `NAME`-on-`PERSON` credit. The error correctly refuses a case the rule cannot express. The problem is the rule, not the check.

Two smaller notes on scope:

* Not TITLE-specific. Under `PERSON` alone the same collision arises with `ALIAS`, `NAME`, `PREFIX`, `SUFFIX`, `USERNAME`. In practice the trigger is "gold uses a depth-2 branch label plus any depth-3 label on that branch". The LOCATION branch escapes here only by luck, since `STREET_ADDRESS` and `ZIP_CODE` both collapse to `ADDRESS` at the default `canonical_depth=3` and land level with `GPE`.
* Scenario 3 had no bug. On `main` it had three passing tests asserting depth-2 golds TP and depth-3 golds FN against depth-2 predictions, exactly what ADR-003 prescribes. It went from correctly scored to un-evaluable.

See my comment on `_project_prediction` for a rule that removes the need for this error, under which `synth_dataset_v2` evaluates with no mapping decision and `TITLE` keeps its own metrics.

---

presidio_evaluator/entity_mapping/mapper.py
```diff
@@ -854,25 +1008,45 @@ def _level(label: str, level: str) -> str:
...
+            path = h_full.canonical_to_branch.get(resolved)
+            if not path:
+                return resolved
+            target_depth = root_annotation_depth
+            if target_depth is None and len(path) >= 2:
+                depths = annotation_depths_by_branch.get(path[1], set())
+                if len(depths) == 1:
+                    target_depth = next(iter(depths))
+            if target_depth is not None and len(path) > target_depth:
+                return path[target_depth - 1]
```
omri374: Suggestion: replace the "one depth per branch" rule with a "deepest gold ancestor" rule.

The projection itself is correct, and I verified it preserves span boundaries. It is a pure per-label rename, so an overshooting `NAME` span becomes an equally overshooting `PERSON` span with IoU semantics untouched (a 2-of-4-token overlap still scores TP at `iou_threshold=0.5` and FP+FN at `0.75`). That was my main worry about moving this into the mapper, and it is handled cleanly.

The issue is that requiring a single annotation depth per branch is stricter than the goal needs, and it forces the blocking error that stops `synth_dataset_v2` (see my comment on the detection block). Here is an alternative that is still a row-independent uniform rename, so it keeps every property above:

> Project each prediction to the deepest gold-vocabulary label that is an ancestor-or-self of it, and leave it unchanged if there is none.

```python
def _project_prediction(label: str) -> str:
    resolved = _resolve(label)
    if resolved is None or resolved == "O":
        return "O"
    path = h_full.canonical_to_branch.get(resolved)
    if not path:
        return resolved
    for ancestor in reversed(path):        # deepest first
        if ancestor in annotation_vocabulary:
            return ancestor
    return resolved
```

where `annotation_vocabulary` is the set of resolved labels actually used in the annotation column. `_resolved_annotation_paths` already computes this.

Because it walks the prediction's own ancestor path, siblings cannot be conflated and the direction stays asymmetric for free. I prototyped it against every case in the current tests plus the ones this PR blocks:

| gold vocabulary | gold | pred | projected | result |
|---|---|---|---|---|
| `{PERSON}` | PERSON | NAME | PERSON | TP, correct |
| `{NAME}` | NAME | PERSON | PERSON | FP+FN, coarser stays penalized |
| `{DATE, TIME}` | TIME | DATE | DATE | FP+FN, siblings stay distinct |
| `{DATE_TIME}` | DATE_TIME | DATE | DATE_TIME | TP, correct |
| `{PII}` | PII | NAME | PII | TP, correct |
| `{TITLE, PERSON}` | TITLE | PERSON | PERSON | FP+FN, correct |
| `{TITLE, PERSON}` | PERSON | NAME | PERSON | TP, correct |
| `{LOCATION, ADDRESS}` | ADDRESS | LOCATION | LOCATION | FP+FN, correct |
| `{LOCATION, ADDRESS}` | LOCATION | GPE | LOCATION | TP, correct |

Rows 6 to 9 are mixed-depth vocabularies that this PR currently rejects. Rows 6 and 7 are `synth_dataset_v2`'s actual situation, and they reproduce ADR-003's Model A and Model B table exactly. Under this rule the blocking error, the notebook boilerplate, and the integration-test boilerplate can all come out. I would keep an INFO issue for mixed gold depths since it is still worth surfacing, it just should not stop evaluation.

---

notebooks/4_Evaluate_Presidio_Analyzer.ipynb
omri374: This is where the problem lies. `synth_dataset_v2.json` annotates both `PERSON` and `TITLE`, so `get_mapped_results_dataframe()` raises `IncompleteMapping` without it.
As mentioned before, let's try to think of ways we don't hide the error but change the behavior so it's possible.

docs/adr/ADR-003-hierarchical-evaluation.md
omri374, line 35: Marking the ADR Superseded and recording the new design here is good practice, thanks for doing it.

One thing that I thought about (and maybe should have thought about before the previous review): there's an example that motivates the whole ADR, a dataset annotating `TITLE` and `PERSON` on the PERSON branch, with a table of expected per-level outcomes for Model A and Model B. That's what we have in `data/synth_dataset_v2.json`. Notebooks 4 and 5 use it. Under the superseding implementation it raises `IncompleteMapping`, so this document now specifies behavior the system cannot produce on the repo's own primary dataset.

I do not think the fix is to rewrite the example. having both TITLE and PERSON is a reasonable annotation scheme (and there a few other examples in the hierarchy) so we should think of a way to allow this behavior if possible.

tests/entity_mapping/test_hierarchical_evaluation.py
omri374: This replaces three tests that asserted correct behavior, not buggy behavior. On `main`, Scenario 3 (`PERSON, LOCATION, NAME, GPE` vs `PERSON, LOCATION, PERSON, LOCATION`) had `test_scenario3_binary_is_perfect`, `test_scenario3_branch_all_tp_because_depth3_maps_to_same_branch`, and `test_scenario3_detailed_depth2_annotations_hit_depth3_annotations_miss`. The last one asserted that depth-2 golds are TPs and depth-3 golds are FNs against depth-2 predictions, which is exactly what ADR-003 prescribes for a coarser model.

So this scenario was unaffected by the bug the PR fixes. It went from correctly scored to un-evaluable, and the coverage it provided (that a coarser model is penalized at detailed but not at branch) is now gone rather than relocated. Under the rule suggested on `_project_prediction` all three original assertions hold unchanged, so this would be a straight revert of this test back to the three.

Also worth adding whichever way this goes: a test pinning the gold vocabulary of `data/synth_dataset_v2.json` (or at least a `PERSON` plus `TITLE` case) through `get_mapped_results_dataframe()`, since that combination is what NB4 and NB5 actually exercise and nothing in the unit suite currently covers it end to end.

---

presidio_evaluator/entity_mapping/mapper.py
omri374, line 126: Nice catch! this fixes a latent bug beyond the stated scope of the PR. The old module-level `_FULL_HIERARCHY` was built from the default hierarchy, so a `CanonicalMapper` constructed with a custom hierarchy silently got wrong branch lookups in issue detection and level projection. I verified a custom taxonomy now resolves correctly end to end. Worth its own line in the CHANGELOG under Bug Fixes, since it is independently useful and unrelated to the granularity change.

presidio_evaluator/evaluation/base_evaluator.py
omri374, line 288: Minor: `SpanEvaluator.calculate_score_on_df()` has no `allow_generic_entities` parameter, so this is silently swallowed by its `**kwargs` on all three calls. It works today because the span evaluator never had generic-entity forgiveness, but a reader would reasonably assume the flag does something for spans, and any future tightening of `**kwargs` turns it into a `TypeError`.

Either accept and ignore it explicitly in `SpanEvaluator` with a docstring note, or add a short comment here saying it only affects token scoring.

Gating this is the right call regardless of where projection ends up living, by the way. `gold=PERSON, pred=PII` cannot be projected downward, so those rows reach the evaluator mismatched and the legacy forgiveness would otherwise flip them into TPs at the detailed level.

---

omri374 (commented):
Thanks! This is a tricky one, because we do have some cases where we have multiple entities on the same branch (like PERSON and TITLE on the synth_dataset) so I suggested some way to fix it but feel free to suggest otherwise.

omri374 (approved):

PR #202, opened 2026-09-07
feat: carry gold span identity into the evaluator via annotation_span_id

omri374:
Builds on #191. Where #191 keeps collapsed labels distinguishable through merge keys (finest-grained label per token), this carries exact entity *instance* identity into the evaluator, so gold span boundaries are recovered rather than inferred from label runs.

## Approach

`span_to_tag` is the code that aligns spans to tokens, so it is also the source of identity: with `return_span_ids=True` it returns, alongside the per-token labels, the index of the span that produced each label (None for `O`). Ids refer to the caller's original span order and are threaded through overlap resolution, so a span split by a higher-score overlap keeps one identity, and a token's label and id cannot disagree by construction.

`InputSample` stores the ids as `span_ids` when tags are created from spans (`create_tags_from_span=True`, the path all JSON-loaded datasets take), and `get_tags` exposes them behind the same opt-in flag. `BaseModel.predict_dataset` carries them into an `annotation_span_id` column with a fixed 6-column schema: the column is always present, all None for samples whose tags were provided directly.

`SpanEvaluator._merge_key_column` prefers the span-id column over the merge-key column when it carries values for the sentence being processed, and `_merge_keys_for` delegates to it instead of duplicating the resolution. Everything downstream (run-splitting in `_create_spans`, the identity check in `_merge_adjacent_spans`) is unchanged; ids flow through the same plumbing as merge keys. `CanonicalMapper` needs no change: its level projections copy the DataFrame and only rewrite the label columns, so the id column passes through every level untouched.

## What this fixes beyond #191

Merge keys distinguish entity types, not instances, so two adjacent entities of the same type were still merged, the limitation #191's description calls out as out of scope:

- `"She visited Paris , London"` — two LOCATION gold spans stayed one span at every level. With ids they stay two.
- Two same-type entities touching with no token between them were one label run and could not be split by merge keys.

## Scope

Gold side only. `batch_predict` returns tags with no instance identity, so prediction spans keep the existing reconstruction (merge keys where present, visible labels otherwise). A `prediction_span_id` constant is defined and the evaluator resolves it symmetrically, so a source that knows its prediction boundaries can opt in later without evaluator changes.

Fallback chain per sentence: span ids where present with values, else merge keys, else visible labels. `test_all_none_id_column_falls_back_to_merge_keys` pins the middle step so an all-None id column cannot disable #191's fix.

## Verification

- 748 passed, 2 skipped (full suite), `ruff` check and format clean on changed files.
- New tests: 5 in `test_span_to_tag.py` (id/tag alignment, None for O, distinct ids for adjacent same-type spans, split span keeps one id, default return unchanged), 3 in `test_base_model.py` (id column values, fixed schema with all-None column, mixed datasets), 5 in `test_span_evaluator.py` (`TestSpanIdColumn`: touching same-type entities split, skip-word-separated same-type entities split, same-id fragments still merge, binary level, merge-key fallback), 1 in `tests/integration/test_data_objects.py` (`span_ids` populated by `create_tags_from_span`).
- The two notebook integration tests assert `predict_dataset`'s exact column list, updated for the new column.

https://claude.ai/code/session_01LJ4tF86qwabSqafCuTMZFM


presidio_evaluator/evaluation/span_evaluator.py
```diff
@@ -327,19 +328,28 @@ def _process_sentence_spans(
...
+        """Name of the column deciding entity identity for a label column.
+
+        Prefers the span-id column (exact entity instances, attached by
+        ``BaseModel.predict_dataset`` from the gold spans) when it is present
+        and carries at least one value for this DataFrame; falls back to the
+        merge-key column (finest-grained label, attached by CanonicalMapper);
+        returns None when neither is available, in which case callers compare
+        the visible label alone.
         """
         from presidio_evaluator.entity_mapping.data_objects import (  # noqa: PLC0415
```
omri374: can this import live on the top?

presidio_evaluator/models/base_model.py
```diff
@@ -55,20 +55,35 @@ def predict_dataset(self, dataset: list[InputSample]) -> pd.DataFrame:
...
+        ``annotation_span_id`` column is attached: the index of the gold span
+        covering each token (None for ``O`` tokens, and for tokens of samples
+        without spans). Downstream span reconstruction uses it to recover exact
+        gold span boundaries instead of inferring them from label runs.
+
         :param dataset: List of InputSample objects (must have tokens and tags set).
-        :return: DataFrame with exactly 5 columns:
+        :return: DataFrame with columns:
             sentence_id, token, annotation, prediction, start_indices
+            [, annotation_span_id]
```
omri374: this should be empty if not computed, not missing. The API should have a fixed set of columns

presidio_evaluator/models/base_model.py
```diff
@@ -55,20 +55,35 @@ def predict_dataset(self, dataset: list[InputSample]) -> pd.DataFrame:
...
         """
+        from presidio_evaluator.entity_mapping.data_objects import (  # noqa: PLC0415
```
omri374: Can this be on top?

presidio_evaluator/models/base_model.py
```diff
@@ -92,6 +107,49 @@ def predict_dataset(self, dataset: list[InputSample]) -> pd.DataFrame:
...
+    def _annotation_span_ids(sample: InputSample) -> list[int | None]:
+        """Index of the gold span covering each token, None for O tokens.
+
+        Ids follow the labels in ``sample.tags``, which span_to_tag already
+        aligned to the tokens: a token labelled O gets None even if a span
+        overlaps it, and when several spans overlap a token the one whose
+        entity type matches the token's label wins, so a token's label and its
+        span id never disagree.
+        """
+        spans = sample.spans or []
```
omri374: if we have spans, do we need to iterate tags?

presidio_evaluator/models/base_model.py
```diff
@@ -92,6 +107,49 @@ def predict_dataset(self, dataset: list[InputSample]) -> pd.DataFrame:
...
             ],
         )
+        if has_spans:
+            # object dtype keeps ids as int/None; a numeric column would turn
+            # None into NaN, and NaN != NaN breaks equality-based grouping.
+            df[ANNOTATION_SPAN_ID] = pd.Series(span_ids, dtype="object")
+        return df
+
+    @staticmethod
+    def _annotation_span_ids(sample: InputSample) -> list[int | None]:
```
omri374: can the span id live in span to tag?

PR #207, opened 2026-09-13
fix: LICENSE resolves to a single entity in the hierarchy (#199)

### LINKED ISSUE #199: LICENSE is defined twice in the entity hierarchy
`LICENSE` appears twice in `presidio_evaluator/entity_mapping/definitions.py`:

- line 166, as a canonical leaf under `EMPLOYMENT`
- line 241, as an alias of `PROFESSIONAL_LICENSE` under `GOVERNMENT_ID`

Nothing says which one wins. `_build_alias_map` writes both and the last write survives, so the alias wins and the leaf becomes unreachable by its own name.

## What it looks like

```python
h = EntityHierarchy()

h.canonicalize("LICENSE")   # 'PROFESSIONAL_LICENSE'
h.to_branch("LICENSE")      # 'EMPLOYMENT'
h.get_branch("LICENSE")     # ['PII', 'GOVERNMENT_ID', 'PROFESSIONAL_LICENSE']
```

Three lookups for one label, two different answers about which branch it belongs to. `to_branch` disagrees with the other two because it checks `canonical_to_branch` on the raw name first, and `LICENSE` is in there as a canonical leaf, so it returns `EMPLOYMENT` and never gets to the alias.

## Why it is worth fixing now

Until recently you could not really run into it, because `add_alias` raised on anything touching `LICENSE`. #192 fixed that guard, correctly. The side effect is that the duplicate is now reachable:

```python
h.add_alias("LICENSE", "PROF_LIC")   # raised before #192, succeeds now

h.canonicalize("PROF_LIC")   # 'LICENSE'               -> EMPLOYMENT
h.canonicalize("LICENSE")    # 'PROFESSIONAL_LICENSE'  -> GOVERNMENT_ID
```

You declare two labels synonyms and get two labels that resolve to different entities in different branches. A dataset annotated `PROF_LIC` and a model predicting `LICENSE` will not match at the exact level or the branch level, and nothing warns.

## Scope

Only this one. I walked every node in the tree and compared the node against where its own name resolves — `LICENSE` is the single case where they disagree.

`DEVICE_IDENTIFIER` has a similar shape (its name resolves to `DEVICE_ID`, an alias on its own child) but both sides stay inside the `DEVICE_IDENTIFIER` branch, so no lookup contradicts another. It is fine as is.

Assuming that LICENSE is government-issued ID, drop the `EMPLOYMENT` leaf and let `LICENSE` resolve to `PROFESSIONAL_LICENSE`.

Found while reviewing #192.

cpruijsen:
`LICENSE` is defined twice in the entity hierarchy: as a leaf under `EMPLOYMENT`, and as an alias of
`GOVERNMENT_ID` > `PROFESSIONAL_LICENSE`. The two lookups disagree, so `canonicalize("LICENSE")`
returns `PROFESSIONAL_LICENSE` while `to_branch("LICENSE")` returns `EMPLOYMENT`, and which one a
caller gets depends on which function it happens to use.

The `EMPLOYMENT` leaf is removed from `presidio_evaluator/entity_mapping/definitions.py`.
`PROFESSIONAL_LICENSE` under `GOVERNMENT_ID` is the better home: it is where the alias already
pointed, and a professional license is an identifier issued by an authority rather than an employment
attribute.

This changes evaluation results for datasets annotated with a `LICENSE` label, which now score against
the `GOVERNMENT_ID` branch. That is the point of the fix rather than a side effect, so it is called
out in the changelog rather than left to be discovered.

`docs/mapping_scenarios.md` drops the `LICENSE` row from its table of duplicate aliases, since it is
no longer one. The remaining rows there are other instances of the same class and are not touched
here.

Fixes #199


omri374 (approved):
Thanks!

